PR #3, opened 2018-03-12
make sure not too much is below the fold

michaelneale:
set the size of the svg to be smaller

screen shot 2018-03-13 at 9 46 15 am


michaelneale:
cc @jstrachan

PR #5, opened 2018-03-15
some doc tweaks for quickstarts

michaelneale:
I thought it may make sense to point to where the quickstart repos are, as people may want to contribute some over time.

cc @jstrachan


rawlingsj:
yeah great idea - thanks @michaelneale

PR #38, opened 2018-04-16
Use last updated date from Git

asgeirf:
Ref. e.g. http://jenkins-x.io/getting-started/create-cluster/

## Before:
screen shot 2018-04-16 at 23 53 34

## After:
screen shot 2018-04-16 at 23 54 51


jstrachan:
thanks!!!

PR #50, opened 2018-04-21
WIP:Document translates to Chinese

LinuxSuRen:


jstrachan:
@LinuxSuRen many thanks, this is awesome! I didn't even realise that Hugo had such great support for multiple languages!

jstrachan:
@LinuxSuRen I guess we need some kinda way to let the user know there's a chinese translation available on the home page somehow? e.g. its live now http://jenkins-x.io/zh/ but from http://jenkins-x.io/ its hard to know its there

LinuxSuRen:
@jstrachan I agree with that. We can put a multi-languages menu on the bottom after I finish the translation.

jstrachan:
@LinuxSuRen sounds great. Maybe we can add something on the top menu bar too to switch between languages?

LinuxSuRen:
Yes, it will let more and more people use Hugo as static website framework.
And maybe we can put some comment about how to contribute i18n translation work.

PR #98, opened 2018-05-03
Add cluster size requirement

ofersadgat:
With fewer than this number of spare vCpus Jenkins will hang on kubernetes failing to schedule the node due to lack of cpu resources available.


michaelneale:
LGTM - this is what @ofersadgat actually had to do to get builds running

PR #99, opened 2018-05-04
Add Chinese translation for create-cluster

LinuxSuRen:


LinuxSuRen:
Hi: @jstrachan 
Here is a problem. Since the description of jx command was auto-generated.
https://jenkins-x.io/commands/jx/
So I can't sync the document between English and Chinese very well.
Maybe we can bring [go-i18n](https://github.com/nicksnyder/go-i18n) into jx project.

PR #160, opened 2018-05-29
don't need to mention path

michaelneale:
and can also use watch


michaelneale:
cc @jstrachan this may not be up to date with the latest stuff

michaelneale:
closing this as needs more work

PR #272, opened 2018-07-02
Docs to explain the opinionated decisions of Jenkins X

rawlingsj:


rawlingsj:
@jstrachan the PR is being rate limited ATM but if you wanted to give this a quick read and when you get 5 mins I'd appreciate it.

rawlingsj:
@LinuxSuRen BTW we might move this about a bit so in case you were going do some translations I'd hold off a bit.

LinuxSuRen:
It's ok, I'm translating faq now.

rawlingsj:
OK great!

michaelneale:
hey @rawlingsj @jstrachan - do you think this could be massaged into a blog post? (I would be happy to) as it is kind of interesting ... (for the Jenkins X audience makes sense as a docs/faq, but thinking of taking it to other audiences). 

Thoughts?

rawlingsj:
Yeah that’s the aim - I did some more work last night to split between Accelerate and our own decisions. Will be tweeting the docs and putting a blog version most likely today.

PR #281, opened 2018-07-04
add oce support in getting-started

hhuiliu:


jstrachan:
note the change to https://github.com/jenkins-x/jx-docs/pull/281/files#diff-876d9993abb6aafb7d5c684aace51b2f will be overwritten - these files are code generated from the output of the `jx` CLI

PR #349, opened 2018-07-21
First four talks about Jenkins-X in German (more to come)

ascheman:
I am giving four talks about Jenkins X at Java user groups and conferences in Germany and Switzerland in September, October, November, and December and would like to see these on the talks list.


ascheman:
Rebasing and submitting new PRs

PR #431, opened 2018-08-20
Update jx_gc.md

pushthat:
No more duplicate on jx gc doc.


jstrachan:
Note this file is auto generated from the source - so we should fix it here... 
https://github.com/jenkins-x/jx/blob/master/pkg/jx/cmd/gc.go#L22

pushthat:
Noted. thank.

PR #540, opened 2018-09-18
Add an installation note about --skip-ingress

mnuttall:
Some Ingress installations are via DaemonSet and do not have an Ingress service or deployment. In those cases, installing with --skip-ingress --domain may be successful.


mnuttall:
@jstrachan This is the followup to the Slack thread https://kubernetes.slack.com/archives/C9MBGQJRH/p1537188295000100

jstrachan:
@mnuttall many thanks!

PR #608, opened 2018-10-01
Typos

jglick:


jglick:
BTW there is some text in the **Maven** section which I could not follow:

> following the #Versioning step mentioned above

Was this supposed to be a link? If so, to what?

jstrachan:
@jglick thanks! I think that `following the #Versioning step mentioned above` looks like a random copy/paste bug ;)

PR #690, opened 2018-11-07
Use a correct URL to the workspace and add a note about signup.

jtnord:
It was not obvious to me where I had to signup as I clicked directly on
the room link.  Also given I already has a slack account for a different
workspace trying to add the new workspace was not obvious because
slack.k8s.io is not really a slack workspace, so changed the second URL
to be the workspace URL and not a vanity one.


jtnord:
image

PR #856, opened 2018-12-18
Update Google Analytics ID

moritzplassnig:
Resulting in a dedicated jenkins-x.io Google Analytics property (separated from jenkins.io)


michaelneale:
nice

PR #893, opened 2019-01-04
…/blob/master/… considered harmful

jglick:
`…/blob/master/…` URLs should not be used in documentation since they could suddenly start referring to something else (or 404s). Press the **Y** key in GitHub to get a permalink, unless you truly wish to link to whatever happens to be at that location today.


LinuxSuRen:
/approve

All links are good.

PR #945, opened 2019-01-15
chore(docs) document the security process

jtnord:
fixes jenkins-x/jx#2684

### DIFF
--- a/content/security.md
+++ b/content/security.md
@@ -1,0 +1,32 @@
+---
+title: Jenkins X Security
+linktitle: Security
+description: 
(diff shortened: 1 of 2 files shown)


content/security.md
```diff
@@ -0,0 +1,32 @@
+---
+title: Jenkins X Security
+linktitle: Security
+description:
```
jtnord: this was all copied from community (except the dates changed) may need tweaking but it looked ok as far as I could tell when run locally.

jtnord:
FYI @daniel-beck

PR #1012, opened 2019-01-24
What is Jenkins X

vfarcic:
Can you please take a look at this @jstrachan? I'm not sure whether that would be useful for https://jenkins-x.io. If it is, can you please let me know where would it make sense to put it?


content/about/what.md
```diff
@@ -0,0 +1,13 @@
+# What Is Jenkins X?
+
+To understand Jenkins X, we need to understand Kubernetes. But, this is not the place where you will learn Kubernetes, but rather extend your knowledge. So, I will assume that you already know what Kubernetes is and how to use it. If you do not, Jenkins X might be overwhelming, and I suggest you learn at least basic Kubernetes features and get a bit of hands-on practice first. I do not expect you to be a Kubernetes ninja and to know everything there is to know about Kubernetes. No one does. As a matter a fact, Jenkins X helps us by simplifying complex processes into concepts that can be adopted quickly and without spending months in trying to figure out "the right way to do stuff." It helps to elevate some of the problems based on the overall complexity of Kubernetes and its ecosystem.
```
rawlingsj: One of the aims of Jenkins X is that you don't need to know Kubernetes and it will help you get going fast and allow you to learn Kubernetes concepts using CI/CD for fast feedback.

I'm wondering if we can reword this as we want anyone who is not familiar to Kubernetes to come straight to Jenkins X?

jstrachan: I'd rather not imply to understand Jenkins X you have to understand k8s - as thats a huge bit of knowledge. I'd rather it be more like Lambda - give us your code + you get CI/CD without needing to know or care about k8s. 

Obviously someone needs to install it and they will need access to k8s - but a developer should be able to treat Jenkins X as a black box that does CI/CD - rather like folks using a Jenkins server without needing to understand Java programming or servlets

vfarcic:
Please let me know if that's the right place to add that doc.

PR #1029, opened 2019-01-28
Update jx_edit_storage.md

mikecirioli:
Update examples to reflect provider naming conventions for storage buckets


jstrachan:
@mikecirioli all the CLI documentation comes from the source code of the commands am afraid - these wiki docs are auto-generated.

I wonder if you could make your changes to the source of the help instead: https://github.com/jenkins-x/jx/blob/master/pkg/jx/cmd/edit_storage.go#L23

Issue #1312, opened 2019-04-02
Update codegen documention


hferentschik:
After the latest changes to code generation in jx we need to update the section in the jx development section covering code generation.

PR #1329, opened 2019-04-03
fix: updating docs around code generation

hferentschik:
- also adding docs around tracing Kube API calls via TRACE_KUBE_API

fixes issue #1312


hferentschik:
tack

PR #1440, opened 2019-04-29
generate redirects

### LINKED ISSUE #1439: hugo redirects are no longer working - and old data is served
since the move from netlify to github pages the redirects are no longer working.

This alas means that there is some old content being serverd.

I tried a quick fix, but the Chinese documentation is using the same aliases without prefixing the URL with `/zh/` and it is a case of last wins so the alias serves a Chinese page if one exists.

jtnord:
given we are no longer using Netlify to serve the pages but GitHub we should put the redirects back.

fixes #1439


jtnord:
/hold
localisation pages trample on the aliases

PR #1554, opened 2019-05-27
Fix broken link for "jx_add_app"

allir:
The link should not point to a "localhost" URL.


MarckK:
Thank you for your PR, @allir! Good spot!

PR #1604, opened 2019-06-07
Fix date jw agenda post

MarckK:


MarckK:
Alyssa's post from yesterday isn't showing on the site in the Jenkins X News section.

I figure this is likely due to how the date was originally configured. Hence this fix. If this is incorrect or there is something else that needs to be changed, please let me know.

PR #1643, opened 2019-06-12
Update serverless-apps.md

sdoxsee:


MarckK:
Good spot. Thank you, @sdoxsee!

PR #1671, opened 2019-06-17
chore: use alibaba svg logo

carlossg:


carlossg:
/assign @jstrachan 

if the rendering supports svg

carlossg:
it works!

PR #1682, opened 2019-06-19
Add JX Pipeline syntax reference doc

abayer:
This is still a work in progress - just wanted to throw up what I've got so far to make sure the structure looks reasonable to everyone.

fixes https://github.com/jenkins-x/jx/issues/4313

/assign @jstrachan 
/assign @wbrefvem 
cc @sharepointoscar


abayer:
Still needs examples, but we can do that separately. Getting the reference up is first priority.

PR #1693, opened 2019-06-21
adding docs for cert-manager issue

haysclark:
documenting a common cert-manager issue effecting kubernetes cluster v1.12 or older

relates to jenkins-x/jx/issues/3134

### DIFF
--- a/content/faq/issues.md
+++ b/content/faq/issues.md
@@ -361,6 +361,56 @@ kail -l  job-name=expose
 
 This will watch for exposecontroller logs and then dump them to the console. Now trigger a promotion pipeline and you should see the output within a minute or so.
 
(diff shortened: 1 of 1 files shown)


haysclark:
/assign @rajdavies

There is no Issue in the _jx-docs_ repo for this PR; however, jenkins-x/jx#3134 (which is already closed) is related.

Currently, new GKE clusters default to Kubernetes 1.12.8-gk3.6 and _cert-manager_ needs a specific label set for clusters v1.12 and older in order for site-wide TLS to work.  The fact that GKE defaults to a version of Kubernetes that has a known issue with _cert-manager_ seems like a good reason to mention this in the FAQ.

PR #1795, opened 2019-08-02
chore: describe how to handle DCO

pmuir:
fixes jenkins-x/jx#4950


MarckK:
Hi @pmuir, Can we add more information for users about signing-off their work? 

For example, the simplest way, especially for first time contributors is to just manually add the sign-off line, or use `git commit -m "Commit Message" -s`. Also, we could tell people how to create git aliases to auto-sign commits.

I'm happy to merge this in and then add some more information regarding the above, does that sound reasonable?

PR #1837, opened 2019-08-19
docs: Viktor's articles and talks

vfarcic:


vfarcic:
It seems that there is no jx pipeline associated with this repo anymore.

vfarcic:
Let's see whether jx works now.

/assign @vfarcic

PR #1869, opened 2019-08-27
fix: lets use relative links

jstrachan:
using relative links so that the links when running the site locally or in a Preview Environment work properly


MarckK:
Really lovely that you have fixed this. Thank you, James 😀

PR #1870, opened 2019-08-27
fix: first spike at docs for lighthouse

jstrachan:


jstrachan:
until the relative links PR merges you can view the preview here: http://jx-docs.jx-jenkins-x-jx-docs-pr-1870.jenkins-x.live/architecture/lighthouse/

PR #1892, opened 2019-09-02
update Azure Kubernetes Service name

cloudmelon:
update Azure Kubernetes Service name, azure container service has been deprecated


cloudmelon:
/assign @ccojocar

Best wishes from the UK
Melony

PR #1910, opened 2019-09-05
docs: Fix broken url

martin-patsov:
Fix broken url of `values.yaml` file in `jenkins-x-platform` repository.


martin-patsov:
I couldn't find if this content is autogenerated from somewhere else so I guess this is the original location of the content.

martin-patsov:
@pmuir "Not mergeable. Should not have needs-ok-to-test label." I think you misspelled `/ok-totest`

PR #1930, opened 2019-09-10
contributing vault content back to OSS docs

jha-cloudbees:
* Enhanced Intro
* Procedure for configuring Public DNS Zones in GKE
* Procedure for configuring TLS settings and vault support in `jx-requirements.yaml`
* Procedure for configuring vault with `jx create vault`.


dennisnewel:
@ccojocar do you happen to know if `jx create domain` works for other k8s flavours than GKE? mainly wondering if we should have a notice saying this is only supported for GKE, or if we can make the text more generic

content/architecture/vault.md
```diff
@@ -18,13 +18,223 @@ toc: true
...
+`192.168.1.100.nip.io`.
+
+6.  The program will ask for an `URLTemplate`. Press `Enter` to
+use the default value.
+
+7.  The program will verify your answers to the previous questions in
+summary and prompt you to approve the Vault creation (default is
+`Yes`).
+
 ## Using Vault with boot
```
ccojocar: @jha-cloudbees I think we can remove this section `Using Vault with boot`. It is already covered by your docs.

ccojocar:
>do you happen to know if jx create domain works for other k8s flavours than GKE? mainly wondering if we should have a notice saying this is only supported for GKE, or if we can make the text more generic

@dennisnewel it only works with GCP:
```
jx create domain -h
Create a Domain in a managed DNS service such as GCP

Available Commands:
  create domain gke Create a managed domain for GKE
```
Actually all instructions only works on GCP. It definitely makes sense to specify this somewhere in the beginning.

content/architecture/vault.md
```diff
@@ -18,13 +18,223 @@ toc: true
...
+email: certifiable@acmecorp.example
+enabled: true
+production: true
+secretStorage: vault
+```
+
+Run `jx boot` for the changes to take effect in your
+environment.
+
+## Creating a Vault
```
ccojocar: I wondering if we should move this section after `Using Vault on the CLI` because `boot` already creates a vault and for most users is more important to know how to use it form CLI after the initial setup. Essentially, how they can update secrets/get secrets from the vault created by boot.

`Create a Vault` is nice to have in case some users want to create additional vaults.

jha-cloudbees:
@jstrachan I hope I got this rebase right 👍

jha-cloudbees:
@ccojocar I believe @dennisnewel reassigned this to himself as part of the handoff to fix it for OSS docs. I just rebased the PR per @jstrachan direction.

dennisnewel:
yeah, i still have it on my todo :) wanted to get the last few bugs on the new site resolved first though.

hopefully i'll pull in both contributions this week

content/architecture/vault.md
dennisnewel: removed the section completely

content/architecture/vault.md
dennisnewel: moved using vault up and updated headers a bit as well. I think we should ideally split the content into "using jx" and "managing jx"

PR #1962, opened 2019-09-18
fixed eks is _free_ bug.

chainsaw-dev:


content/about/decisions.md
```diff
@@ -29,7 +29,8 @@ Kubernetes has won the cloud wars, every major cloud provider now either support
 
 The Kubernetes ecosystem is rich with innovation and with a vibrant, forward thinking, diverse open source community which is inviting only suggests great things for all involved.
 
-Jenkins X strongly recommends using public cloud managed Kubernetes clusters where possible. GKE, AKS and EKS all manage and run your Kubernetes masters for __free__ you pay for the worker resources required to run your applications.  This dramatically reduces risk of installing, upgrading and maintaining your Kubernetes cluster.  
+Jenkins X strongly recommends using public cloud managed Kubernetes clusters where possible. GKE, AKS and EKS all offer managed Kubernetes services, This dramatically reduces risk of installing, upgrading and maintaining your Kubernetes cluster so you can focus on developing awesome code.
```
dennisnewel: super small detail, but would you mind lower-casing the t in "this"? :) or maybe replace it with "which"?

chainsaw-dev: aah yes good catch, i think "which" is better here.

Issue #1884, opened 2019-08-28
Download does not exist


mcasperson:
The instructions on https://jenkins-x.io/getting-started/install/ say to download the file from https://github.com/jenkins-x/jx/releases/download/latest/jx-linux-amd64.tar.gz. This file does not exist.

dennisnewel:
I'll update the link. It shouldn't be a valid link as there's no option to link to a "latest" version on github (as far as i can tell) so you'll need to grab the correct URL from the https://github.com/jenkins-x/jx/releases/ page as mentioned above the link

dennisnewel:
apparently there was a shortcode/command thingy to insert the latest version, so i used that instead. that should hopefully provide a correct link to the latest version

dennisnewel:
this should be fixed now

PR #1963, opened 2019-09-18
fix: using version variable to generate updated urls for downloading jx

dennisnewel:
use the {{< version >}} shortcut/command in the download links to make sure the displayed links are always linking to the latest version.

assuming here that the version from the `config.toml` file is always the latest


dennisnewel:
this is in response to https://github.com/jenkins-x/jx-docs/issues/1884

dennisnewel:
this PR resolves #1884

Issue #1993, opened 2019-09-20
the note shortcode is either not working or just not styles


dennisnewel:
should replace it with the one that comes with docsy called `alert`

dennisnewel:
closing as the fix has been merged

PR #1994, opened 2019-09-20
Remove old admonitions and use docsy default instead

dennisnewel:
replaced `note` with `alert`
replaced `warning` with `alert color="warning"`


dennisnewel:
fixed #1993

PR #1997, opened 2019-09-21
fix: adding spring boot content from CJXD back to oss docs, take 2

jha-cloudbees:
Was told to redo this PR after doing a rebase. I hope I did it right 🤞


jha-cloudbees:
@jstrachan you said I should rebase this PR, hope this means I did it successfully.

PR #2011, opened 2019-09-25
fix: polish the navigation tree

jstrachan:
* moved more components to the `reference/components` section 
* added Boot into the `setup` section
* added a couple more FAQs


daniel-beck:
Neither security nor sponsors seems to make sense below "Contributing to Jenkins X".

PR #2060, opened 2019-10-03
docs: edit shell commands for contributing

MarckK:


content/en/docs/contributing/documentation/_index.md
```diff
@@ -69,29 +69,29 @@ Now open your fork repository on GitHub and copy the remote url of your fork. Yo
 Then go back to your terminal, `cd` to where you would like to place your local copy of the `jx-docs` repo, and then clone your fork.
 
 ```shell
-$ git clone --recurse-submodules --depth 1 git@github.com:<YOUR_USERNAME>/jx-docs.git
-$ cd jx-docs
+git clone --recurse-submodules --depth 1 git@github.com:<YOUR_USERNAME>/jx-docs.git
```
hferentschik: ooi, why would you remove the '$'? I think it is common practice to have '$' or '>' in this case to symbolise the prompt.

This is in particular handy when you have a command + its output. In this case it makes clear which lines are input and which are output returned from the command.

PR #2061, opened 2019-10-04
docs: remove $ from shell command

MarckK:
check: locally created changes. user.email set to GH 'private' user email.


MarckK:
@hferentschik do you mind if I merge this, given that we will likely move the command-line prompt, `$` to be a symbol outside of the code snippet?

PR #2064, opened 2019-10-04
chore: Add 'How do I inject Vault secrets...'

silphid:
This is based on my experimentations and understanding, please correct me if I'm wrong! ;)

### DIFF
--- a/content/en/docs/using-jx/faq/_index.md
+++ b/content/en/docs/using-jx/faq/_index.md
@@ -254,3 +254,41 @@ preview:
   env:
     PASSWORD: vault:path/to/mysecret:password
 ```
+
+## How do I inject a Vault secret via a Kubernetes Secret?
+
+When you inject secrets directly into environment variables, they appear in Deployment yaml as plain text, which is not advisable. It is recommended to rather inject them into a Secret yaml that will itself be mounted as environment variables.
+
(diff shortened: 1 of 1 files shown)


dennisnewel:
looks good @silphid !

i'm wondering if it would be better to have a dedicated page under Using Jenkins X -> Developing just about using Vault? We can always expand more on it later

Knappek:
Awesome @silphid , this documentation is needed. Looks good, but after enabling vault in staging/production, it is also possible to inject secrets like you do for the preview environment by simply using the `vault: path/to/mysecret` syntax in your apps repo in `chart/myapp/values.yaml`. Hence you don't need to distinguish between preview and staging/production.

silphid:
@dennisnewel There is already this page about Vault: https://jenkins-x.io/docs/reference/components/vault/  Should I add my documentation at the end?

dennisnewel:
@silphid i think this content is better under the "Using...". Maybe just call it "Using Vault"?

One thing i tried to do with new structure is to split things up between managing and using, with the hope that it'd be easier to find what you're looking for depending on how you use jx; the content itself just hasn't been updated to that yet.

silphid:
Thanks @Knappek, indeed it also works to inject directly into the app's values.yaml. I added a note in that regard! :)

PR #2010, opened 2019-09-24
docs: update link to Contribute page in different language versions

szemek:
https://jenkins-x.io/docs/contribution-guidelines/
returns
image

This pull request fixes references to Contribute page and it is pointing to https://jenkins-x.io/docs/contributing/

PR #2066, opened 2019-10-06
fix : Modified path create-cluster to /docs/getting-started/setup/create-cluster

chandanikumari:
Author: Chandani Kumari, Committer: Chandani Kumari <chandanikumari92@gmail.com>


MarckK:
Awesome work, @chandanikumari!

We have a  中文 translation of the site. You can view it as an option under the languages section (the tab labelled `English` in the top nav bar). 

It would be fantastic of you to make the proposed link change to the 中文 versions of the site, similar to PR #2010.

chandanikumari:
@MarckK Its done and PR is raised :)

PR #2070, opened 2019-10-07
fix: Adds missing word

mauromorales:
My guess is that the desired expression is "out of the box" and not "of the box". I'm not a native english speaker so if this is not the case just feel free to close this PR. Thanks!


hferentschik:
@mauromorales, thanks for the PR. You will need to make sure that your commits are signed off (git `-S ...`) or you sign them with your ssh key.

hferentschik:
I guess you want to add yourself to the OWNERS file as well.

PR #2123, opened 2019-10-20
fix: link to api-conventions updated

butuzov:
Updated link to api conventions following it's move by
kubernetes/community@06207d23

### DIFF
--- a/content/en/docs/reference/api.md
+++ b/content/en/docs/reference/api.md
@@ -93,7 +93,7 @@ Kubernetes meta/v1.ObjectMeta
 <td>
 <em>(Optional)</em>
 <p>Standard object&rsquo;s metadata.
-More info: <a href="https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata">https://git.k8s.io/community/contributors/devel/api-conventions.md#metadata</a></p>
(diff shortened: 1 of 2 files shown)


dennisnewel:
@jstrachan this stuff is generated i think? do you know where to change the source?

pmuir:
They are generated from the source code that we import. The only way to fix this to update to a kubernetes release that has the links fixed in the source.

butuzov:
Is it possible that something broken in the pipeline process?

For example, I have tracked this line "https://github.com/jenkins-x/jenkins-x-versions/blob/master/jx/scripts/update-jx-website.sh#L52", but  I can't even install `get-crd-api-referrence-docs` to test it... is it possible that apidocs.md isn't actually updating or something?

jstrachan:
I wonder if these URLs come from the specific version of the k8s API we build jx against?

butuzov:
I believe they come from https://github.com/jenkins-x/jx/blob/d8c3e032e903b54c39b5c9d775dd09a6b3eb6af0/Makefile.codegen#L60:L73, but I can't build, but I wasn't able to perform step `install-refdocs`, as some of the conflicts happened during go modules pulling. didn't go further...

hferentschik:
Closing this pull request, since the fix needs to happen on the code generation side.

PR #2161, opened 2019-10-31
feat: translated to Spanish Getting started section

mmorejon:


mmorejon:
Vamos por más! Estoy ahora en la sección Concepts!

content/es/docs/getting-started/demos-talks-posts/articles.md
```diff
@@ -0,0 +1,52 @@
+---
```
vfarcic: I don't think it makes sense to translate this page, unless someone wants to commit to keeping it translated continuously.

content/es/docs/getting-started/demos-talks-posts/talks.md
```diff
@@ -0,0 +1,80 @@
+---
```
vfarcic: This is a live page that is being updated continuously. Can someone commit to keeping it always up to date? If not, it's better to leave it in english.

content/es/docs/getting-started/demos-talks-posts/articles.md
sharepointoscar: Yes, agreed.  @mmorejon maybe these can stay the way they are?  Also less work maintaining the translation if we don't bother translating these...

content/es/docs/getting-started/demos-talks-posts/talks.md
sharepointoscar: I recommend not translating this particular page.

mmorejon: Sounds good for me. One question, whats happens with the `es` folder structure?
- should I remove the file completely? Will not appear in the navigation bar.
- should I keep some reference to the generated page in the `en` folder?

content/es/docs/getting-started/demos-talks-posts/articles.md
mmorejon: Good. This case is similar to the question asked in the following conversation.

content/es/docs/getting-started/demos-talks-posts/talks.md
mmorejon: @vfarcic @sharepointoscar ?

PR #2168, opened 2019-11-02
fix: correct ingress example indenting

joostvdg:
The ingress configuration example's indenting is not showing correctly. Some parts of the file were incorrectly indented, but for some reason there entire thing is not showing any indenting. So this might not fix it either...

### DIFF
--- a/content/en/docs/reference/components/vault.md
+++ b/content/en/docs/reference/components/vault.md
@@ -215,9 +215,9 @@ ingress:
   externalDNS: true
   namespaceSubDomain: -jx.
   tls:
-email: certifiable@acmecorp.example
-enabled: true
-production: true
+    email: certifiable@acmecorp.example
+    enabled: true
+    production: true
 secretStorage: vault
 ```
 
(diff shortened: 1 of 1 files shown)


dennisnewel:
@joostvdg guess that didn't fully do it 🤔 http://jx-docs.jx-jenkins-x-jx-docs-pr-2168.jenkins-x.live/docs/reference/components/vault/

trying it locally, it looks like the whole block might need to be indented to start under the second bullet (i.e. six spaces until the backticks start)

joostvdg:
@dennisnewel the indenting is good now, but not sure if this is how we want to do this

dennisnewel:
hmm..yeah, the other bullets seem to have gone a bit out of place now?  i'll give it a shot...see if i can somehow convince it to do proper indenting. I've had issues on other pages with getting the tripple back tick to work with lists...

dennisnewel:
ended up restructing most of the page :) it didn't flow very well, and with smaller chunks of lists, the indenting worked out better as well.

let me know what you think

joostvdg:
Looks good to me

PR #2148, opened 2019-10-25
docs: add rich event metadata

karentobo:
Resolves #2133 by adding rich event metadata to one blog entry for Jenkins World and one event for FOSDEM.

I didn't use a JSON-LD file, but just added the metadata in two formats:

- The FOSDEM event supplies all values as metadata without affecting the visible content of our page.
- The Jenkins World event mixes the event data in with the visible content.

I'm a tester with automation skills and not at all a UI developer, so I'd love feedback to make this easier to copy/paste to other events.

PR #2171, opened 2019-11-02
docs: add rich events readme

karentobo:
As a follow-on to Dennis's suggestion in #2148, this adds a readme with the link to Google's rich event docs and two sample events.


dennisnewel:
Thanks Karen! i might look into creating that as a shortcode this week if i have time. So far, I don't see any events show up in Google, but i think we're waiting for them to crawl the site again, which could take some time

/ok-to-test
/lgtm
/approve

PR #2237, opened 2019-11-22
fix: remove out-dated tutorial for installing jx on gke

dennisnewel:


sharepointoscar:
I am wondering if it makes more sense to actually update the tutorial rather than deleting it?  I feel folks want tutorials...

dennisnewel:
fair point @sharepointoscar - that would be the best solution. it's not good that we're confusing people though, so i prefer to take it down until we have a new, improved version :)

sharepointoscar:
Let's do it.  We can come up for a new one soon.

PR #2270, opened 2019-12-03
chore: adding sponsors section to the landing page

### LINKED ISSUE #2265: Move sponsors to the front page and add products built with JX
I know we've only got CloudBees as a sponsor right now, but i think it'd be good to show that there's someone backing the project and it might encourage others to join in as well :) 

Along side sponsors, we should show the product built with (or projects using) Jenkins X. There's a lot of work happening outside of JX that feeds a lot of fixes and improvements back into the project, so i think it's fair to show these off as well

dennisnewel:
fixes #2265 

instead of moving the page, i've just added a new section for sponsors and their products on the landing page.

this PR also includes the CDF logo + link in the navbar as suggested in #2265


dennisnewel:
@jstrachan @MarckK wdyt?

PR #2276, opened 2019-12-05
fix: moving favicons to get the robot back instead of the default theme icon

dennisnewel:


dennisnewel:
favicons are back on the preview environment, so assuming it'll be good once it hits production.

/approve

PR #2299, opened 2019-12-19
feat: Existing Vault Instance note

sharepointoscar:
Added note on no supportability for existing Vault instance, though open issue exists and will be available sometime soon!


vfarcic:
I would suggest that if you mention that there is an open issue, it might be a good idea to provide a link to it. That way, those interested can subscribe to receive notifications of the progress.

sharepointoscar:
The repo where the issue is logged is private, otherwise I would have.

Sent from my iPhone X

On Dec 19, 2019, at 1:33 PM, Viktor Farcic <notifications@github.com> wrote:

﻿

I would suggest that if you mention that there is an open issue, it might be a good idea to provide a link to it. That way, those interested can subscribe to receive notifications of the progress.

PR #2300, opened 2019-12-19
build: bumping hugo to 0.65.3. also switching to new htmltest image

dennisnewel:
Note that this also includes changes to a large number of files where the Alert or Pageinfo short codes were used. The format for short codes seems to have changed (or behaviour have changed a bit) so changed them to standard format as shown in https://gohugo.io/templates/shortcode-templates/


dennisnewel:
finally a green build. html-proofer is working as a replacement to htmltest and hugo-0.65.3 builds the site correctly.

going to approve

/approve

PR #2307, opened 2020-01-03
fix: Change docs to refer to correct jx-requirements.yml filename

msembinelli:
Add note explaining why it must be .yml and not .yaml

Helps explain [#5141](https://github.com/jenkins-x/jx/issues/5141) and clear up confusion with documentation

The requirements loader in the jx cmd code will only load jx-requirements.yml and NOT jx-requirements.yaml. I think that should also be changed to allow for both extensions, however the easy fix right now is to clear up the confusion with the docs first.


msembinelli:
Any reason why this is failing pr-build? I don't have permissions to view the dashboard obviously

PR #2314, opened 2020-01-07
docs: brew, use one-liner where brew install does it all

maelvls:
From https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap#installing:

> ### Installing
> If it’s on GitHub, users can install any of your formulae with `brew install user/repo/formula`. Homebrew will automatically add your `github.com/user/homebrew-repo` tap before installing the formula. `user/repo/formula` points to the `github.com/user/homebrew-repo/**/formula.rb` file here.
>
> If they want to get your tap without installing any formula at the same time, users can add it with the `brew tap user/repo` command.

Can't tell when that was introduced, but it makes the `brew` installation a one-liner 😁

### DIFF
--- a/content/en/docs/getting-started/setup/install/_index.md
+++ b/content/en/docs/getting-started/setup/install/_index.md
@@ -17,8 +17,7 @@ Pick the most suitable instructions for your operating system:
     On a Mac you can use [brew](https://brew.sh/):
 
 ```sh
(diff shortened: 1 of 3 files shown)


rawlingsj:
@maelvls nice!!! Thanks for this, I'm just going to edit the PR title so that the semantic pull request check passes

PR #2328, opened 2020-01-09
docs: add missing Fosdem Devroom managers

olblak:


MarckK:
Awesome, thank you for adding these names, @olblak!

PR #2340, opened 2020-01-14
docs: small tweaks for install instructions

michaelneale:
hey @MarckK @sharepointoscar I added a link to oscars step by step, and mention CJXD as a "stable distro" (suggested by Ethan). 

Oscar may be working that into a guide, and I think @rawlingsj had some other ideas - but please take a look to see if it makes things clearer (also a typo with kubernetes in one spot). 

I think this all needs more work, both in CLI for UX but also docs to explain what is needed (I can't quite work out where), so if this gets in the way of that, please close it.

### DIFF
--- a/content/en/docs/getting-started/_index.md
+++ b/content/en/docs/getting-started/_index.md
@@ -3,7 +3,7 @@ title: "Getting Started"
 linkTitle: "Getting Started"
 weight: 2
 description: >
-  How to get quickly up and running with Jenkins X
+  How to get up and running with Jenkins X
 aliases:
   - /getting-started
 ---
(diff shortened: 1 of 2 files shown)


tomfrenken:
Use the prefix `docs:` in the title so the semantic bot doesn't block the PR

michaelneale:
@tomfrenken the  title of the PR or the commit? sorry couldn't quite work out from that semantic PR stuff (is new to me!)

tomfrenken:
Both works, I didn't understand that part at first as well 😅 - but it appears you did it right

michaelneale:
thanks @tomfrenken - it seems it needs a space after the docs like `docs: ` for it to pick it up correctly.

PR #2369, opened 2020-01-21
docs: add new categories with jx edit storage

tomfrenken:
Previously, I wasn't aware that you could also create new categories with `jx edit storage` as there was no documentation about it, I also improved a few readability related things.


dennisnewel:
looks good @tomfrenken! one small heads up for next time is that the commands are actually generated from code, so changing them via a PR won't last (they get overwritten with new releases of jx)

/lgtm

PR #2383, opened 2020-01-25
docs: remove cloud shell instructions

MarckK:


MarckK:
Git version conflict with Cloud Shell persists, therefore taking out the Cloud Shell instructions for creating a cluster.

PR #2434, opened 2020-02-17
fix: command to to download the jx binary

MarckK:
Fix is for Create Cluster docs.


MarckK:
These docs are under 'deprecated', but still visible to end user, and if you search for 'cloud shell' on docs site this is the top result.

PR #2440, opened 2020-02-18
docs: Add Vault section to EKS boot document

deanesmith:
This PR addresses item 2 described in https://github.com/jenkins-x/jx/issues/6732.  Vault on EKS using Boot,


deanesmith:
> This looks solid. I specifically like the section on providing existing Vault instance information given the parameters needed. I wonder if it makes sense adding some instructions on creating the artifacts for an existing Vault, or do we imply user should know (which is totally fine)?

There will be quite a bit more content provided specifically for this as these concerns are properly decoupled from `jx boot`.  Ref https://github.com/jenkins-x/jx/issues/6734.  Other than the CloudFormation template referenced in the document, I don't have any other samples.  Feel free to further this!

PR #2495, opened 2020-03-05
feat: Add instructions for debugging jx boot pipeline

mrageh:
Debugging a boot pipeline with dlv is something we sometimes may want to
do, this commit adds instructions for how to do that with dlv.

### DIFF
--- a/content/en/docs/contributing/code/_index.md
+++ b/content/en/docs/contributing/code/_index.md
@@ -584,6 +584,31 @@ If you want to debug using `jx` with `stdin` to test out terminal interaction, y
 	dlv --listen=:2345 --headless=true --api-version=2 attach SomePID
 	```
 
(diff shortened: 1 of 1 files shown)


content/en/docs/contributing/code/_index.md
```diff
@@ -584,6 +584,31 @@ If you want to debug using `jx` with `stdin` to test out terminal interaction, y
 	dlv --listen=:2345 --headless=true --api-version=2 attach SomePID
 	```
 
+### Debugging jx boot pipeline
+
+If you want to debug a pipeline, for example `install-vault` pipeline you'll have to change the args and command in `jenkins-x.yml` file in the boot config directory:
```
hferentschik: You mean debugging local execution, not in cluster, right? Might be worth clarifying.

PR #2555, opened 2020-03-20
docs: updating content around Terraform usage

hferentschik:
- adding documentation around Terraform modules and how to use them
- restructuring getting started navigation
- cleaning up Markdown syntax

fixes #2485

PR #2570, opened 2020-03-24
[wip] restructure of docs landing pages

jha-cloudbees:
restructure of main docs landing, create cluster, and quickstart pages with new sequential ordered cards/buttons

### DIFF
--- a/content/en/docs/_index.md
+++ b/content/en/docs/_index.md
@@ -33,7 +33,7 @@ your operating system.
 
 [More installation options... ](/docs/getting-started/setup/install/)
 
-### Step 2: Configure your Kubernetes environment
+### Step 2: Provision your Kubernetes Cluster
 
 Once you have installed the binary on your workstation, next you can configure your kubernetes environment and install Jenkins X.
 
--- a/content/en/docs/_index.md
+++ b/content/en/docs/_index.md
@@ -1,5 +1,5 @@
 ---
-title: Documentation
+title: Jenkins X Documentation
 linkTitle: Documentation
 weight: 20
 menu:
@@ -8,3 +8,71 @@ menu:
 aliases:
   - /documentation
 ---
+
+Jenkins X is an open source project that provides Kubernetes-based CI/CD
+applications featuring pipeline automation, built-in GitOps, and preview
+environments to help teams collaborate and accelerate their at any scale. [Read
+More...](/docs/overview/)
+
+## Get started with Jenkins X in 3 easy steps!
+
+
+### Step 1: Install the Jenkins X Binary
+
+The first step in using Jenkins X is installing the Jenkins X binary (`jx`) for
+your operating system.
+
+<div style="text-align:center">
+<a class="btn btn-lg btn-secondary mr-3 mb-4" href="/docs/getting-started/setup/install/#linux">
+		Install Linux <i class="fab fa-linux ml-2 "></i>
+    </a> 
+<a class="btn btn-lg btn-dark mr-3 mb-4" href="/docs/getting-started/setup/install/#macos">
+		Install macOS <i class="fab fa-apple ml-2 "></i>
+	</a>
+</div>
+
+[More installation options... ](/docs/getting-started/setup/install/)
+
+### Step 2: Configure your Kubernetes environment
+
+Once you have installed the binary on your workstation, next you can configure your kubernetes environment and install Jenkins X.
+
+<div class="row">
+  <div class="col-sm-4">
+    <div class="card">
+      <div class="card-body">
+        <h5 class="card-title">Google Clusters</h5>
+        <p class="card-text">How to create a kubernetes cluster on Google Cloud Platform (GCP)</p>
+        <a href="/docs/getting-started/setup/create-cluster/google/" class="btn btn-primary">Google <i class="fab fa-google ml-2 "></i></a>
+      </div>
+    </div>
+  </div>
+  <div class="col-sm-4">
+    <div class="card">
+      <div class="card-body">
+        <h5 class="card-title">Amazon Clusters</h5>
+        <p class="card-text">How to create a kubernetes cluster on Amazon Web Services (AWS)</p>
+        <a href="/docs/getting-started/setup/create-cluster/amazon/" class="btn btn-secondary">AWS <i class="fab fa-aws ml-2 "></i></a>
+      </div>
+    </div>
+  </div>
+  <div class="col-sm-4">
+    <div class="card">
+      <div class="card-body">
+        <h5 class="card-title">Azure Clusters</h5>
+        <p class="card-text">How to create a kubernetes cluster on Microsoft Azure platform.</p>
+        <a href="/docs/getting-started/setup/create-cluster/azure/" class="btn btn-dark">Azure <i class="fab fa-microsoft ml-2 "></i></a>
+      </div>
+    </div>
+  </div>
+</div>
+
+### Step 3: Install Jenkins X on your cluster
+
+Once you have configured your kubernetes environment, you can finally Install Jenkins X and start using it!
+
+<div style="text-align:center">
+<a class="btn btn-lg btn-info mr-3 mb-4" href="/docs/getting-started/setup/boot/">
+		Install Jenkins X <i class="fas fa-arrow-alt-circle-right ml-2"></i>
+	</a>
+</div>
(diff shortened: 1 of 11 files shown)


content/en/docs/_index.md
```diff
@@ -8,3 +8,71 @@ menu:
...
+		Install Linux <i class="fab fa-linux ml-2 "></i>
+    </a> 
+<a class="btn btn-lg btn-dark mr-3 mb-4" href="/docs/getting-started/setup/install/#macos">
+		Install macOS <i class="fab fa-apple ml-2 "></i>
+	</a>
+</div>
+
+[More installation options... ](/docs/getting-started/setup/install/)
+
+### Step 2: Configure your Kubernetes environment
```
sharepointoscar: Probably change this header to  **Step 2: Provision your Kubernetes Cluster**

content/en/docs/_index.md
```diff
@@ -8,3 +8,71 @@ menu:
...
+  </div>
+  <div class="col-sm-4">
+    <div class="card">
+      <div class="card-body">
+        <h5 class="card-title">Azure Clusters</h5>
+        <p class="card-text">How to create a kubernetes cluster on Microsoft Azure platform.</p>
+        <a href="/docs/getting-started/setup/create-cluster/azure/" class="btn btn-dark">Azure <i class="fab fa-microsoft ml-2 "></i></a>
+      </div>
+    </div>
+  </div>
```
sharepointoscar: For the cloud logos, can we use the actual logo?  I think it resonates very well with folks and it's almost second nature in terms of recognition.  

I believe we can use the svg versions.

content/en/docs/_index.md
```diff
@@ -8,3 +8,71 @@ menu:
...
+
+### Step 3: Install Jenkins X on your cluster
+
+Once you have configured your kubernetes environment, you can finally Install Jenkins X and start using it!
+
+<div style="text-align:center">
+<a class="btn btn-lg btn-info mr-3 mb-4" href="/docs/getting-started/setup/boot/">
+		Install Jenkins X <i class="fas fa-arrow-alt-circle-right ml-2"></i>
+	</a>
+</div>
```
sharepointoscar: We probably don't want to show any further content below this section, like it shows now.  Which means we can try and use more real estate on the page to outline the three steps.

hferentschik:
I am a bit confused here. The landing page is looking very much like a "Getting Started" section which I think is good by the way. At the same time, the actual "Getting Started" menu item is still there and looks very different. I am a bit lost about the bigger picture here.

content/en/docs/labs/boot/getting-started/run.md
```diff
@@ -20,6 +20,14 @@ jxl boot run
 
 This will use helm to install the boot Job and tail the log of the pod so you can see the boot job run. It looks like the boot process is running locally on your laptop but really it is all running inside a Pod inside Kubernetes.
 
+Once this has finished you are ready to import or create a quicstart.
```
hferentschik: quickstart

content/en/docs/_index.md
```diff
@@ -8,3 +8,71 @@ menu:
...
+
+[More installation options... ](/docs/getting-started/setup/install/)
+
+### Step 2: Provision your Kubernetes Cluster
+
+Once you have installed the binary on your workstation, next you want to provision your kubernetes cluster.  Choose your cloud provider.
+
+<div class="row">
```
hferentschik: Do we really need so much custom HTML? Have we explored all options to get this decently laid out and rendered with Markdown and a minimal amount of HTML?

content/en/docs/getting-started/first-project/_index.md
```diff
@@ -5,3 +5,62 @@ weight: 2
...
+.bg-black {background: black;}
+.bg-red {background: red;}
+</style>
+
+If you're just starting your development process with Jenkins X and want to start with some tested, supported project templates that you can setup quickly, you can create a **Quickstart**!
+
+Quickstarts are pre-made applications you can start a project from, instead of starting from scratch.
+
+
+<div class="row">
```
hferentschik: Are these boxes really worth getting in so much HTML? IMO it really destroys the concept of using something like Hugo or AsciiDoctor.

content/en/docs/labs/boot/getting-started/cloud/gcloud.md
```diff
@@ -44,6 +44,13 @@ cd cloud-resources/gcloud
 ./create_cluster.sh
 ```
 
+<nav>
+  <ul class="pagination">
```
hferentschik: Should we really start messing with the built-in navigation? What is your concern here? That the page is so long and the top breadcrumb is gone? Maybe better in this case to split the page?

content/en/docs/labs/boot/getting-started/cloud/gcloud.md
```diff
@@ -60,3 +67,10 @@ This way we'll list out all the [gcloud](https://cloud.google.com/sdk/gcloud) co
 <div class="highlight">
 <pre style="color:#f8f8f2;background-color:#272822;-moz-tab-size:4;-o-tab-size:4;tab-size:4"><code class="language-bash hljs" data-lang="bash">{{<gcp-create-resources>}}</code></pre>
 </div>
+
+<nav>
```
hferentschik: Same as above.

content/en/docs/labs/jxl.md
```diff
@@ -17,6 +17,9 @@ e.g. `jxl` assumes:
 * helm 3 is used
 * we use lighthouse and tekton for CI/CD
 
+notes:
```
hferentschik: don't we want to use the same style for all notes? Something like:

```
{{< pageinfo >}}
**NOTE** 
....
{{< pageinfo >}}
```

hferentschik:
@jha-cloudbees unfortunately, there is lots of overlap with https://github.com/jenkins-x/jx-docs/pull/2555. We will need to synchronize how we want to move forward. Eventually one will have to rebase on the other.

I like that there is a new landing page. The current state where it basically did just mirror the left side navigation did not make sense. Now we effectively start with "Getting Started". What confuses me about this is that it conflicts now with the existing "Getting Started". How will that be consolidated? 

Also, the new landing page refers to content I am in the process of removing. I could imagine a situation where we use your new landing page, but then link to the new content I am creating.

Have you seen my attempts to make the "Getting Started" read better - http://jx-docs.jx-jenkins-x-jx-docs-pr-2555.jenkins-x.live/docs/getting-started/? My approach was to streamline and improve the current content/flow. I think if done correctly Hugo's default navigation can work as well. 

I am very concerned about seeing so much custom HTML. Is that really required and does using ut make things so much better to justify its use? I am also concerned that the HTML is tied to the structure HUGO generates. That might change, right? Would it not be better to have something like \<box\> around the quickstart examples and the style that in CSS? In my opinion the actual docs page should stay in almost pure Markdown.

hferentschik:
@dennisnewel any thoughts? ^^

dennisnewel:
agree with @hferentschik on the use of custom HTML. This should really be created as short codes using the Hugo templating. The generated HTML could change without notice, so it's not a good idea to rely on that.
If the theme needs to be changes, remember to copy-paste the file you want to change to /layout/partials from /themes/layout/partials and change the one in /layout/partials. This is because the other files are technically not part of the site but a copy of the theme which will be replaced next time the theme is updated.

In terms of overlapping work, i think getting Hardy's new content in first would be best. The new landing page can then be updated to reference that instead of old content

hferentschik:
@jha-cloudbees, just talked with @sharepointoscar and we thought it might be easiest to close this PR and let @sharepointoscar rebase this work on master. 

As it stands now, this pull request would at the very least need a rebase, but also some work related to the custom HTML. WDYT?

sharepointoscar:
> @jha-cloudbees, just talked with @sharepointoscar and we thought it might be easiest to close this PR and let @sharepointoscar rebase this work on master.
> 
> As it stands now, this pull request would at the very least need a rebase, but also some work related to the custom HTML. WDYT?

@jha-cloudbees I think it makes sense.  It will be a cleaner approach and since our work overlaps it makes sense for me to start from the existing master branch as well.

hferentschik:
> it makes sense for me to start from the existing master branch as well.

+1, maybe even try to break up the work. Let's, for example, update the theme first and get it merged (if that's possible). That way further changes are easier to review and comprehend.

sharepointoscar:
closing as per our overall plan to move forward.

sharepointoscar:
Closing this PR as per our internal discussion.

PR #2585, opened 2020-03-30
docs: add labs external dns docs

rawlingsj:


hferentschik:
@rawlingsj, what is the state of this PR. Just wondering. I am thinking you might have put it on hold because the preview environments were not working.

PR #2595, opened 2020-04-02
chore: updating and restructuring the contributing to docs pages

hferentschik:
Content is pretty much still the same. I tried to update commands which were not correct anymore (used spell checker for example), but mainly I tried to break up the massive single page into more palatable smaller pieces.


hferentschik:
@MarckK, @sharepointoscar, @jha-cloudbees - an update and restructuring of the "contribute to the documentation" pages. WDYT?

MarckK:
Awesome! Thank you, @hferentschik!

/lgtm

PR #2674, opened 2020-04-28
fix: disabling site translations

### LINKED ISSUE #2673: Disable translations
After the latest re-organisation of content, the docs translations differ a lot in structure from the main documentation. IMO we should stop deploying them until we get an update contributed. 

This will also give us some time to address the issue that our search functionality searches across all languages regardless of the currently selected language.

hferentschik:
fixes #2673

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -16,27 +16,28 @@ enableGitInfo = true
 contentDir = "content/en"
 defaultContentLanguage = "en"
 defaultContentLanguageInSubdir = false
+disableLanguages = ["es", "zh"]
 
 # Useful when translating.
 enableMissingTranslationPlaceholders = true
 
 [languages]
-[languages.en]
-title = "Jenkins X"
-description = "Jenkins X documentation site"
-languageName ="English"
-# Weight used for sorting.
-weight = 1
-[languages.zh]
-languageCode = "zh-CN"
-weight = 2
-languageName = "中文"
-contentDir = "content/zh"
-[languages.es]
-languageCode = "es-ES"
-weight = 3
-languageName = "Español"
-contentDir = "content/es"
+	[languages.en]
+		title = "Jenkins X"
+		description = "Jenkins X documentation site"
+		languageName ="English"
+		# Weight used for sorting.
+		weight = 1
+	[languages.zh]
+		languageCode = "zh-CN"
+		weight = 2
+		languageName = "中文"
+		contentDir = "content/zh"
+	[languages.es]
+		languageCode = "es-ES"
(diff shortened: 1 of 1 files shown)


MarckK:
Granted the translations are now out of sync structurally and possibly otherwise with the english documentation, but does that mean they no longer have any value and should be removed from the site (even if temporarily)?

jha-cloudbees:
Agree with @MarckK, temporarily removing support to clean up the restructure would be fine for now, with caveat that we can re-include once there are cycles for translator to refresh the translated content if possible.

hferentschik:
@mmorejon and @LinuxSuRen, I think you are the main contributors to the translations. Would you be ok with temporarily disabling them? There are two reasons for that:

1. The content got quite heavily moved around and is not mirrored in the translations. 
1. We are having issues with the search functionality of the site. The search is across all translations and when searching for a given term one quite often ends up with content in a different language. We are working on a fix for that, but that might take some time, especially since we are in the process of figuring out who initially setup the account with Algolia. 

WDYT?

If you guys want to address the first point, that would be absolutely awesome. We are trying to resolve the second asap.

mmorejon:
Hi @hferentschik, I agree to temporarily delete the translation. I hope to continue supporting you with the documentation as soon as possible.

LinuxSuRen:
I am not sure disabling translations is a good approach. That will be a big break for those users. About the second issue, I believe there must a better solution.

hferentschik:
> I am not sure disabling translations is a good approach. That will be a big break for those users. 

Sure, it is not optional. On the other hand, the experience is getting worse and worse. The moving of pages is one thing, but we hope to get some major changes to the content in as well. Unless the translations happen at the same time I fear there might be too much divergence. 

> About the second issue, I believe there must a better solution.

Absolutely. This needs to be fixed. Algolia has the capabilities to properly separate the indices. This requires some time on the index configuration as well as how we integrate with the Algolia search API. We are working on this, but having issues with getting access to Algolia. Right now the experience when searching is really bad and that impacts the default docs. 

Maybe we can speed up the process of getting the search fixed. That's really what bothers me most. Whether out of date content in the translations is better than no content is then a different question.

LinuxSuRen:
I can help with Chinese translation from the structure. Further more, we might well have a chance to improve the documentation well under the Doc SIG #2675

MarckK:
@deanesmith is handling access to Algolia search and hopefully that will be resolved soon. 👍

hferentschik:
Closing this pull request, since I found a way to handle translations.

PR #2717, opened 2020-04-30
chore: adding labs warning to all labs related documentation

vfarcic:
This PR adds the warning from the initial labs page to all the labs pages. That way, people will know that it is experimental and that the current labs are closed no matter whether they navigate through the left-hand navigation, or stumble upon one of the pages through, let's say, Google search.

### DIFF
--- a/content/en/docs/labs/boot/_index.md
+++ b/content/en/docs/labs/boot/_index.md
@@ -7,7 +7,17 @@ weight: 7
 aliases:
   - /artwork
 ---
+{{% alert %}}
+**NOTE: This current experiment is now closed. The work done and feedback we have received will be used to enhance Jenkins X in future versions**
 
+**This code should not be used in production, or be adopted for usage.  It should only be used to provide feedback to the Labs team.**
+
+Thank you for your participation,
+
+-Labs
+
+
+{{% /alert %}}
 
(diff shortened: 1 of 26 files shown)


sharepointoscar:
@vfarcic and/or @hferentschik is this still relevant? I believe we have the verbiage on the Labs landing page already.

hferentschik:
@sharepointoscar 

> is this still relevant? I believe we have the verbiage on the Labs landing page already.

Right, the warning is once on the landing page, but all other pages don't contain the warning. So if you don't read it on the landing page or if you even not come in via the landing page you might miss it. My guess is that the attempt here is to make sure that on each Labs related page it becomes clear that it was an experiment. 

Not sure really what to do. Happy to merge I guess. @ethangj wdyt? 

Personally I am wondering whether this documentation should be even there at all.

hferentschik:
@vfarcic, technically is there a chance you can update the commit message to follow the conventional [commit message format](https://www.conventionalcommits.org/en/v1.0.0/)? Something like "chore: adding labs warning to all labs related documentation".

vfarcic:
My bad... Sorry for that... It should be OK now.

ethangj:
My opinion is that if the docs are gonna be live (which they already are), we need to be over the top with warnings so there's no chance at confusion.

PR #2730, opened 2020-05-04
fix: configuring docsearch to target correct index facet

### LINKED ISSUE #2729: Take currently selected language into account when searching
When using the Algolia search field, currently all languages are targeted. The Algolia index already contains a language facet which should be targeted depending on the selected language.

hferentschik:
includes some minor cleanup and formatting as well

fixes #2729


hferentschik:
@ethangj, thanks. So much better with proper search context :-)

PR #2762, opened 2020-05-11
fix: setting baseURL to fix broken sitemap

ethangj:
Because the config.toml doesn't have a proper base URL set, the sitemap is not generating full URLs and thus is invalid in Google. This change will let the sitemap generate full URLs and allow Google to crawl the sitemap.

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -1,4 +1,4 @@
-baseURL = "/"
+baseURL = "https://jenkins-x.io"
 title = "Jenkins X - Cloud Native CI/CD Built On Kubernetes"
 
 enableRobotsTXT = true


ethangj:
@hferentschik why the hold?

hferentschik:
> why the hold?

I am wondering whether there were side effects for the local site. Wanted to double-check. Also, I was looking at something else. It feels wrong to set this deployment-specific setting in `config.toml`. I would think that would be more of an environment setting. I wanted to check whether it is not already done in `jenkins-x.yml` by setting `HUGO_BASEURL` or by passing the `--baseURL` variable.  Apparently not.

ethangj:
Updating config.toml seems to be the recommended solution: https://discourse.gohugo.io/t/google-search-console-reports-sitemap-xml-as-invalid/10474/11

Also didn't impact local dev for me, but there is a way to spin up the local server to override the setting if it does encounter that problem.

hferentschik:
> Updating config.toml seems to be the recommended solution

It is not so much changing `config.toml`, but setting the base URL. The latter can happen in three ways, in `config.toml`, by setting the command line flag and by setting an environment variable. 

For the preview we already set an explicit https://github.com/jenkins-x/jx-docs/blob/master/jenkins-x.yml#L123. I think it would be best to do the same for the release. At least things are consistent then and not handled differently in various places.

hferentschik:
@ethangj, my vote goes to set this in `jenkins-x.yml`, but if you really want I can take the hold off as well.

ethangj:
@hferentschik If that's something you could help define in the pipeline, that would be great. Otherwise would prefer to merge this then refactor later since this isn't causing any issues + solves an important blocking issue.

hferentschik:
> If that's something you could help define in the pipeline, that would be great.

Here we go - https://github.com/jenkins-x/jx-docs/pull/2764

ethangj:
Confirmed solved by 2764

PR #2768, opened 2020-05-12
docs: add eks ingress section to faqs

### LINKED ISSUE #2767: Add common EKS issue to FAQ section
Related to this: 
https://github.com/jenkins-x/jx/issues/7150

ankitm123:

fixes #2767

### DIFF
--- a/content/en/docs/resources/faq/issues.md
+++ b/content/en/docs/resources/faq/issues.md
@@ -460,6 +460,31 @@ Delete the local git `jenkins-x-boot-config` repository.
 
 That should leave your Kubernetes provider and your local environment in a clean state.
 
+## No External IP assigned to the ingress in EKS
+
+If you have never created an Elastic Load Balancer (ELB) in AWS, then the `jx boot` command will fail to assign an ip to the ingress controller.
+
+In this case, the output from `jx boot` might look like this:
+```sh
+error: failed to discover the Ingress domain: getting a domain for ingress service kube-system/jxing-nginx-ingress-controller: Timed out after 5m0s, last error: %!s(<nil>)
+```
+
+To verify this is the case, run:
+```sh
+kubectl get svc --all-namespaces
+```
+
+If you see `pending` for external IP for `jxing-nginx-ingress-controller`
+```sh
+NAMESPACE     NAME                                  TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                      AGE
+.....
+kube-system   jxing-nginx-ingress-controller        LoadBalancer   X.X.X.X         <pending>     80:32632/TCP,443:32036/TCP   4h45m
+...
+```
+then create an ELB manually outside of EKS, and tear it down.
+
+After that, run `jx boot` again.
+
 ## Other issues
 
 Please [let us know](https://github.com/jenkins-x/jx/issues/new) and see if we can help? Good luck!


content/en/docs/faq/issues.md
```diff
@@ -460,6 +460,32 @@ Delete the local git `jenkins-x-boot-config` repository.
 
 That should leave your Kubernetes provider and your local environment in a clean state.
 
+## No External IP assigned to the ingress in EKS
+
+If you have never created an Elastic Load Balancer (ELB) in AWS, then the `jx boot` command will fail to assign an ip
```
hferentschik, line 465: Do you mind putting this into a single line? We try to aim for line per sentence since that makes diffing changes easier.

hferentschik: @ankitm123 - Is this specific to using a custom domain or even an issue when using nip.io? I have not seen this myself, just wondering. Might be worth clarifying. 

Do you know what the underlying reason is for that? Is there a bug somewhere on our side or is this an AWS thing?

@dgozalo - do you know more about this?

ankitm123: From what I know, this is on AWS.

ankitm123: > Do you mind putting this into a single line? We try to aim for line per sentence since that makes diffing changes easier.

Done 👍

ankitm123: > @ankitm123 - Is this specific to using a custom domain or even an issue when using nip.io? I have not seen this myself, just wondering. Might be worth clarifying.

I did not use custom domain, but I have seen others using custom domain run into this issue. Example: https://github.com/jenkins-x/jx/issues/7150
 
> Do you know what the underlying reason is for that? Is there a bug somewhere on our side or is this an AWS thing?
> 

This is an AWS issue. There is some relevant information here: https://github.com/terraform-aws-modules/terraform-aws-eks/issues/183#issuecomment-435315065

ankitm123: Also this from AWS: https://eksworkshop.com/beginner/050_deploy/servicerole/, weird ...

ankitm123: Actually think about it a bit more and looking at some other issues (https://github.com/terraform-aws-modules/terraform-aws-eks/issues/87), not sure how much of a value would it bring to include the additional burden of creating a service link role for elb. It can just be documented that a service link role for ELB should be created before provisioning EKS. Even AWS does the same thing in their eks workshop.

hferentschik: > A fix would be to add the creation of this role in the jx aws eks module. I can test it out in my test aws account, and if it works, I can create a PR. @hferentschik wdyt?

+1,  I always prefer to fix the root cause :-)

hferentschik: @ankitm123, there is also some discussion around missing tags!? Quite confusing everything.

hferentschik: @ankitm123, what do you think we should do here now?

ankitm123: I think we should document it, this is a a case that should not be handled by the eks module anyways. Even AWS in their EKS workshops makes it a pre-req for people to create a service role for ELB before creating an EKS cluster in a new AWS account. I feel adding the ability to create this role in jx terraform module just creates more maintenance burden, given the fact this only happens for new AWS accounts, with no ELBs created before.  

Even the upstream EKS module added it, and then reverted it back (https://github.com/terraform-aws-modules/terraform-aws-eks/issues/87#issuecomment-461119382). All in all, seems like documenting this is much better than making a code change. @hferentschik, thoughts? I think AWS might fix it in the future, seems more like a cloud provider issue ...

ankitm123:
There was a merge conflict after another PR was merged, so resolved and pushed the changes again.

ankitm123:
The needs ok to test label should be removed ...

PR #2776, opened 2020-05-13
docs: create event calendar page

MarckK:


hferentschik:
@MarckK 

> create event calendar page

what's the intend. For what I can tell this adds a new "Event Calendar" item to the "Community" docs. But what is the difference to "Events and Webinars" which already contains the same calendar, right?

MarckK:
@hferentschik, Yes, all the calendars point to the same JX calendar, so all are the same. 
We may reduce the number of calendars proliferating on the pages, but for now I've just mirrored all the places we were giving office hours information.

This calendar functions as the way to submit events to the calendar. 

Other pages could simply link to this one in the future, including from governance repos -- for the Docs SiG, for example.

hferentschik:
I see. Personally I have a hard time distinguishing between the purpose of these calendars. 

> This calendar functions as the way to submit events to the calendar.

In this case should this not go somewhere under "how to contribute to the docs"? I guess what confuses me is the same calendar under two very similar sounding top levels menu items. Could one not just add the note about how to add an event to the calendar to the "Events and Webinars" page as well. Why does one have to duplicate the actual calendar?

MarckK:
@hferentschik, you're right:  there is too much unnecessary duplication of the calendar (and office hours info).

I like the clarity of doing one express thing per page, so I like this calendar page -- users can quickly go there to look at the calendar (easily found in search, etc), with no extraneous info, and has instructions to add events to the calendar.

I think the Events and Webinars page / section needs a rethink. Just listing previous talks on [Demos and Talks page ](https://jenkins-x.io/community/events/talksdemos/past-talks-and-demos/)is not that helpful to anyone and it's a visual overload of words. We could list the talks that have slides or videos, and links to those resources. This would then function as a 'Resources to learn more about JX page'.  And, in fact, we have a [Resources Section](https://jenkins-x.io/docs/resources/demos-talks-posts/).

It's good to reduce duplication, but I would reduce it in this case by taking out [Demos and Talks](https://jenkins-x.io/community/events/talksdemos/past-talks-and-demos/) page and cleaning up [Events and Webinars](https://jenkins-x.io/community/events/) page. And, in fact, if it is reduced down to a series of links, then could be a section on main Community landing page:

- Go here to see calendar and add events to it
- Go here to find out about office hours
- Go here to see previous talks and demos

hferentschik:
Sounds good to me. You want t still merge this or continue working on it?

MarckK:
This can be merged.  Thank you, @hferentschik.

I'll clean up Events & Webinars section in separate PR.

PR #2780, opened 2020-05-14
docs: clean up community pages

MarckK:
This page was added here by mistake in the restructuring of docs.

We could have a sponsorship page **somewhere else** in the docs, but 
1) the page should look better than the current page does 
2) it is perhaps unnecessary given we have sponsorship info writ large on the homepage:

Screenshot 2020-05-14 at 11 55 48


hferentschik:
@MarckK, I have nothing against removing this, but I think it was added there intentionally. Leaving this to @ethangj or @sharepointoscar to answer. 

Conceptually I think is nothing wrong with placing this under "community". I agree though that it is a bit of duplication.

ethangj:
I'm fine removing it.

PR #2785, opened 2020-05-14
docs: add events section to landing page; remove events directory

MarckK:


MarckK:
@hferentschik as discussed, removing events directory and adding events section on community landing page.

This is better, cleaner. 

There is an odd README.md file that was under the events directory. It is not event specific. Rather than delete it, I've just moved it further up the tree in the community directory. That isn't where it belongs, and I'll raise an issue about it.

hferentschik:
> This is better, cleaner.

+1 Much much better. Thanks for this cleanup. I think it removes a lot of confusion.

> There is an odd README.md file that was under the events directory. It is not event specific. Rather than delete it, I've just moved it further up the tree in the community directory. That isn't where it belongs, and I'll raise an issue about it.

Right, I have a feeling it is obsolete and we might as well delete it. If there is value in it, we maybe should add the content somewhere to "contributing to the docs" part or otherwise out of the rendered content. I would expect that all content and markdown files under the content directory belong to the rendered site. 

Another quick win imo is to move the "Apps Framework" out of Community. This is documentation about the Apps framework and should go into the main documentation section. Maybe under "Guides and Tutorials" and have a new sub-heading "Extending Jenkins X" under which we place the Apps Framework. WDYT? Do you mind addressing this as well? Otherwise, I might have a go at it.

MarckK:
@hferentschik, Completely agree with your comments. Security page should also be moved into its own section in the documentation imo.

Reading over "Apps Framework", there seem to be a number of TODOs. I like your idea of how to place this within the main documentation. You are welcome to have a go at it. 👍

PR #2818, opened 2020-05-27
fix: restoring /docs redirect

ethangj:
The /docs direct went missing, causing a 404 on a common page URL. Restoring the alias here.

### DIFF
--- a/content/en/docs/getting-started/_index.html
+++ b/content/en/docs/getting-started/_index.html
@@ -5,7 +5,8 @@ description="How to get up and running with Jenkins X"
 
 aliases = [
     "/getting-started/",
-    "/docs/getting-started"
+    "/docs/getting-started",
+    "/docs"
 ]
 
 type= "docs"


content/en/docs/getting-started/_index.html
```diff
@@ -5,7 +5,8 @@
 
 aliases = [
     "/getting-started/",
-    "/docs/getting-started"
+    "/docs/getting-started",
+    "/docs"
```
jha-cloudbees, line 9:
```suggestion
    "/docs"
    "/docs/"
```

content/en/docs/getting-started/_index.html
```diff
@@ -5,7 +5,8 @@
 
 aliases = [
     "/getting-started/",
-    "/docs/getting-started"
+    "/docs/getting-started",
```
hferentschik, line 8: Is this really needed? That's really just a redirect to itself, right? Luckily it seems to get ignored. @ethangj have you tried without?

ethangj:
@jha-cloudbees did you try both combinations to see if both were necessary?

ethangj:
@hferentschik can you try your lgtm again

PR #2823, opened 2020-05-29
docs: removed Google Analytics and replaced it with Google Tag Manager

Dgiordano33:
- Removed Google Analytics
- Added Google Tag Manager

Both were in the <head>. Google Tag Manager is a faster, more flexible method of pushing tags like Google Analytics on the web.


hferentschik:
@Dgiordano33 can you describe more the intention of this change? I think @ethangj has recently been using Google Analytics for this site, but I am not sure. @ethangj wdyt?

PR #2830, opened 2020-06-05
feat: first draft of OpenShift documentation

dgozalo:
First draft of OpenShift documentation.

### DIFF
--- a/content/en/docs/install-setup/installing/boot/clouds/openshift.md
+++ b/content/en/docs/install-setup/installing/boot/clouds/openshift.md
@@ -1,0 +1,202 @@
+---
+title: OpenShift
+linktitle: OpenShift
+description: Using Boot On OpenShift
+date: 2020-05-06
+publishdate: 2020-05-06
+lastmod: 2020-05-06
+weight: 200
+---
+
+{{% alert title="Warning" color="warning" %}}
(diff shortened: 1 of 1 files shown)


content/en/docs/install-setup/installing/boot/clouds/openshift.md
```diff
@@ -0,0 +1,196 @@
+---
+title: OpenShift
+linktitle: OpenShift
+description: Using Boot On OpenShift
+date: 2020-05-06
+publishdate: 2020-05-06
+lastmod: 2020-05-06
+weight: 200
+---
+
```
hferentschik, line 10: Maybe we could add an infobox (something like this here https://jenkins-x.io/docs/install-setup/create-cluster/gke/#cluster-provisioning) which puts into perspective what using Jenkins X on OpenShift means.

AFAIU, this will help you to run `jx boot` on an OpenShift cluster, but there won't be any persistent storage nor Vault. Not sure about things like DNS and TLS.

I would love to be able to manage the users expectations properly here.

dgozalo: Good idea, added a warning on that, should show in the preview.

hferentschik:
sweet

/lgtm

PR #2831, opened 2020-06-05
docs: explain how users can specify custom buildpacks

mrageh:
The `jx edit buildpack` command only works on clusters created with install rather than boot. That's because the `jx edit buildpack` command edits the TeamSettings in the environment CRD. Each time the boot pipeline runs, the TeamSettings are overridden. By the default values found in `jxboot-resources` Helm Chart as part of the boot config - https://github.com/jenkins-x/jenkins-x-boot-config/tree/master/env/jxboot-resources.

We've introduced a new way to specify custom buildpacks in boot based clusters [here](jenkins-x/jx#7172).

### DIFF
--- a/content/en/docs/reference/commands/jx_edit_buildpack.md
+++ b/content/en/docs/reference/commands/jx_edit_buildpack.md
@@ -5,7 +5,24 @@ slug: jx_edit_buildpack
 url: /commands/jx_edit_buildpack/
 description: list of jx commands
 ---
+## Customising buildpacks for boot based clusters
+
+If you want to specify a custom buildpack for a boot based cluster. Then you
+have to add the following fields to the jx-requirements.yml file before running
+boot. All the fields have to be added together otherwise jx won't be able to
+figure out where to get the buildpack from.
+
+```yaml
+buildPacks:
+  buildPackLibrary:
+    name: "Test name"
+    gitURL: "github.com/jx-user/test-repo"
+    gitRef: "master"
+```
+
 ## jx edit buildpack
+
+**This command only works for clusters created with install rather then boot.**
 
 Edits the build pack configuration for your team
 
@@ -22,13 +39,13 @@ jx edit buildpack [flags]
 ```
   # Edit the build pack configuration for your team, picking the build pack you wish to use from the available
   jx edit buildpack
-  
+
   # to switch to classic workloads for your team
   jx edit buildpack -n classic-workloads
-  
+
   # to switch to kubernetes workloads for your team
   jx edit buildpack -n kubernetes-workloads
-  
+
   For more documentation see: [https://jenkins-x.io/architecture/build-packs/](https://jenkins-x.io/architecture/build-packs/)
 ```
 


hferentschik:
@mrageh - fwiw, the build is failing due to some broken links:

```
- ./tmp-website/docs/guides/extending-**/introduction/index.html
  *  linking to internal hash #creating-new-build-packs that does not exist (line 0)
     <a href="/docs/create-project/build-packs/#creating-new-build-packs">creating build packs</a>
- ./tmp-website/docs/reference/components/cloud-native-jenkins/index.html
  *  linking to internal hash #creating-new-build-packs that does not exist (line 0)
     <a href="/docs/create-project/build-packs/#creating-new-build-packs">documentation on creating your own build packs</a>
htmlproofer 3.15.3 | Error:  HTML-Proofer found 2 failures!
```

ethangj:
@mrageh do you need any help with this one?

content/en/docs/reference/commands/jx_edit_buildpack.md
```diff
@@ -5,8 +5,25 @@ slug: jx_edit_buildpack
 url: /commands/jx_edit_buildpack/
 description: list of jx commands
 ---
+## Customising buildpacks for boot based clusters
+
+If you want to specify a custom buildpack for a boot based cluster. Then you
```
hferentschik, line 10: Sorry, but you cannot do this ;-)
If you wanted to make this particular change you need to update _edit_buildpack_ in the _jx_repo. The Markdown for the command reference is generated from the jx repo. Cobra the CLI framework we use can generate this Markdown from the Go files which we then include here. 

You see pull request doing this for example here - https://github.com/jenkins-x/jx-docs/pull/2807

It does actually happen when the version stream is released - https://github.com/jenkins-x/jenkins-x-versions/blob/master/jx/scripts/update-jx-website.sh

Anyways, modifying this manually will lead to merge conflicts once a generated pull request comes in again.

I would have thought that his page - https://jenkins-x.io/docs/create-project/build-packs/ - might be the place to start. Probably needs a bit of an overhaul, but in the end, that's where it should go.

content/en/docs/reference/commands/jx_edit_buildpack.md
```diff
@@ -5,8 +5,25 @@ slug: jx_edit_buildpack
...
+buildPacks:
+  buildPackLibrary:
+    name: "Test name"
+    gitURL: "github.com/jx-user/test-repo"
+    gitRef: "master"
+```
+
 ## jx edit buildpack
 
+**This command only works for clusters created with install rather then boot.**
```
hferentschik, line 25: That actually should be added in the jx codebase. This should be shown to the user when he uses the CLI.

hferentschik:
@mrageh - I think we can close this pull request. See https://github.com/jenkins-x/jx-docs/pull/2831#discussion_r437910157

The part we want to be changed in the CLI command needs to be changed in the jx codebase and the build pack documentation in general should probably go here https://jenkins-x.io/docs/create-project/build-packs/

PR #2841, opened 2020-06-09
feat: updating Vault documentation

### LINKED ISSUE #2825: Update Vault documentation
Most information around Vault is currently in https://jenkins-x.io/docs/reference/components/vault/.

ToDo:

* Shorten intro to Vault
* Describe installation of an external Vault instance
* Describe accessing secrets
* Describe secret rotation

Also integrate with Vault section in the Boot document.

hferentschik:
fixes #2825


hferentschik:
Old - https://jenkins-x.io/docs/reference/components/vault/
New - http://jx-docs.jx-jenkins-x-jx-docs-pr-2841.jenkins-x.live/docs/reference/components/vault/

content/en/docs/reference/components/vault.md
```diff
@@ -4,106 +4,56 @@ linktitle: Vault
...
-Vaults are provisioned in kubernetes using `vault-operator`, an
-open-source Kubernetes controller installed when Vault is configured
-during cluster creation and Jenkins X installation on the cluster.
-
-# Using Vault on the CLI
+## Using Vault on the CLI
 
 First you need to download an install the [safe](https://github.com/starkandwayne/safe) CLI for Vault.
 
 Once you have installed [safe](https://github.com/starkandwayne/safe) you can run:
```
ankitm123, line 40: Why safe? Why not the bank-vault cli, since we use bank vault anyways? Is safe better?

hferentschik:
Vault specific:

* http://jx-docs.jx-jenkins-x-jx-docs-pr-2841.jenkins-x.live/docs/install-setup/installing/boot/secrets/#vault
* http://jx-docs.jx-jenkins-x-jx-docs-pr-2841.jenkins-x.live/docs/reference/components/vault/

hferentschik:
Would like to find a better style for admonitions, but that's a different pull request.

ethangj:
I don't like the premise of a bunch of JX boot specific pages, because the user shouldn't care necessarily about jx boot, they care about the things they're trying to get done - i.e., I like the existing upgrade JX page being where it's at, not as a child of JX boot.

Similarly, I'd see Secrets management as a subset of installing JX, not of jx boot. JX boot isn't a task for the user so I wouldn't make it a parent item of anything.

ethangj:
Maybe the JX Boot page you've added should become "Install JX", with ingress, secrets etc all becoming "Modify Ingress", "Modify secrets" etc as children of it? Then leave Upgrade as a peer of Install, not as a child of it.

hferentschik:
> I don't like the premise of a bunch of JX boot specific pages, because the user shouldn't care necessarily about jx boot,

Boot is at the heart of things. It is the one thing users need to know (at least for now). This is how you bootstrap and later modify/maintain the cluster via Gitops. Having a proper explanation of what Boot is and what the various sections of `jx-requirements.yml` mean is imo essential.

Note, that the boot page already existed, I just broke it up. It was already long, adding all the content around Vault would have made it even longer. I prefer smaller pieces of information.

> they care about the things they're trying to get done - i.e., I like the existing upgrade JX page being where it's at, not as a child of JX boot.

I did not move the existing upgrade page. It still exists where it is (which imo is the wrong place, since installing != upgrading). There is a duplicating upgrade section in the Boot documentation which I split out. I agree that as a follow up we should unify and de-duplicate the upgrade docs. Where it then lives is another question. 

> Similarly, I'd see Secrets management as a subset of installing JX, not of jx boot. JX boot isn't a task for the user so I wouldn't make it a parent item of anything.

You cannot really understand Jenkins X and the dev repo atm, without understanding Jenkins X Boot. Not quite sure what you want to do. The Boot documentation is already part of "Install Jenkins X", right? Do you want to get rid of the "Boot" concept altogether in this section? If so, do you have some concrete suggestion on how to spin this? 

We also say installation of Jenkins X is a three-step process:

1. Create cluster
1. Install `jx` binary
1. Run `jx boot`

If you don't want to go into Boot, what it is and how it works, how are you going to put these steps then? 

Somewhere we need to explain Jenkins X Boot and `jx-requirements.yml`, one way or another. I am happy to move it to somewhere else if you have any good suggestions. For now I kept the existing location.

hferentschik:
> Similarly, I'd see Secrets management as a subset of installing JX, not of jx boot.

Installation and configuration of secret management is part of installing Jenkins X. You need to make a decision at boot time what secret engine to use, local vs vault. That goes into `jx-requirements.yml` which is one of the most central configuration files atm in Jenkins X. 

The actual secret management, how you can update the secrets and how you can rotate them is under the "components" page of Vault. Eventually, I would see more component pages fleshed out and the whole "Components" moved out of its current location. That what's eventually becomes "tasks", rotate secrets for example.

> JX boot isn't a task for the user so I wouldn't make it a parent item of anything

Sorry, I disagree on this one. `jx boot` is one of the key tasks. It is a sub-tasks of the overarching install of Jenkins X, but it is also the key tasks in managing Jenkins X via GitOps.

hferentschik:
> Maybe the JX Boot page you've added should become "Install JX", 

what do you do then with the other pages under "Install Jenkins X"? So far we said that 'Boot' is a sub-task of the installation.

> with ingress, secrets etc all becoming "Modify Ingress", "Modify secrets" etc as children of it?

If you are concerned about the titles of the subheading, we can try to find better ones. At the stage of installing, it is not about "modifying anything so, it is about "configuring" (you are setting it up for the first time). So then it would be more "Configure Ingress", "Configure secrets", etc. Whether 'configure' needs to be repeated on each menu item is then imo arguable. That said, happy to change link titles if that makes it any better.

 > Then leave Upgrade as a peer of Install, not as a child of it.

I did not move upgrade. It is still where it is. My change is just highlighting an existing duplication. There are several other things which are not ideal, however, the pull request is already very big. I thought I reached a point where it could be merged.

@ethangj, if I get rid of the "upgrade" duplication would that help?

hferentschik:
@ethangj - I removed the duplicated "upgrade" information under Boot. It was a bit outdated anyway. It also covers auto-upgrades, but I think this is something we want to rethink anyways. The same for Velero.

For what it's worth, the same type of duplication exists between "Configure SCM" under "Install and Setup" and the git provider documentation under Boot. Mind you imo there is a huge gap in the documentation under "Configure SCM". It goes straight into `jx-requirements.yml` which you would not know what it is unless you know what Jenkins X Boot is and how it works. 

So this is what we have:

>Jenkins X has support for github.com for source code management by default. To enable support for other providers to work with Jenkins X, configure the jx-requirements.yml file according to your preferred SCM provider.

In the current flow as a user not knowing anything about Jenkins X, I would wonder:

- What is ` jx-requirements.yml`? 
- Where do I find this file?
- Is it enough to just edit this file? (you need to trigger the Boot Pipeline)

I think we should discuss how we want to move forward with this, but I don't think this should stop this PR from going in.

content/en/docs/install-setup/installing/create-cluster/_index.md
```diff
@@ -1,7 +1,7 @@
 ---
 title: Create cluster
 linktitle: Create cluster
-description: How to prepare a Kubernetes cluster for Jnekins X
+description: How to prepare a Kubernetes cluster for Jenkins X
```
hferentschik, line 4: @garethjevans, thanks :-)

ethangj:
@hferentschik what if the current "Install JX" page became "Download JX" instead (since it's really not about installing in the normal sense" and the "Run Boot" page becomes "Install JX"? All the children would be fine then

Then, we'd just add language to the top that says "You install JX by using Boot" etc?

hferentschik:
> what if the current "Install JX" page became "Download JX" instead (since it's really not about installing in the normal sense" and the "Run Boot" page becomes "Install JX"? All the children would be fine then

what do you do with the cluster creation & Terraform part? I do believe, at least for now, the structure this PR shows makes sense.

hferentschik:
> Then, we'd just add language to the top that says "You install JX by using Boot" etc?

BTW, you don't install JX you install Jenkins X. I think we need to be concise in language. JX (even in capital letters) is to easily confused with `jx`.

ethangj:
@hferentschik I think you are assuming too much context on the user's part, what may be technically correct is not necessarily what the mental model of a new user looking for information would be. We need to meet a user's mental model and then map it to our concepts, not the inverse.

Install tends to mean "set the thing up", download tends to mean "getting access to the thing". And on the sidebar, you want actions, not functions. The user is going to be scanning for "How do I install JX?" not "how do I JX boot?"

My proposal would be:

- Renaming this to "Download Jenkins X": http://jx-docs.jx-jenkins-x-jx-docs-pr-2841.jenkins-x.live/docs/install-setup/installing/install-binary/

- Renaming this to "Installing Jenkins X": http://jx-docs.jx-jenkins-x-jx-docs-pr-2841.jenkins-x.live/docs/install-setup/installing/boot/

- Add a subhead that says "Jenkins X is installed, configured and managed using the command `jx boot` "

- Rename this to "Setting Up SCMs"

- For secrets/storage/ingress/artifacts/webhooks, add the word "Configuring" to the menu item

- Move secrets down so that it's part of the grouping with the others

also roping in @jha-cloudbees and @vfarcic for feedback

content/en/docs/install-setup/installing/boot/how-it-works.md
```diff
@@ -68,11 +69,12 @@ env/
...
+then you can omit the `prow:` indentation in the `env/prow/values.yaml` file to make the YAML you create/edit smaller and simpler.
+It also means longer term we can generate JSON schema files for each `env/$app/values.yaml` files so that we can support better editor/IDE integration.
 
 #### values.tmpl.yaml templates
 
-When using `jx step helm apply` we now allow `values.tmpl.yaml` files to use go/helm templates just like `templates/foo.yaml` files support inside helm charts so that we can generate value/secret strings which can use templating to compose things from smaller secret values. e.g. creating a maven `settings.xml` file or docker `config.json` which includes many user/passwords for different registries.
+When using `jx step helm apply` we now allow `values.tmpl.yaml` files to use Go/Helm templates just like `templates/foo.yaml` files support inside Helm charts so that we can generate value/secret strings which can use templating to compose things from smaller secret values. e.g. creating a maven `settings.xml` file or docker `config.json` which includes many user/passwords for different registries.
```
jha-cloudbees, line 77:
```suggestion
When using `jx step helm apply` we now allow `values.tmpl.yaml` files to use Go/Helm templates just like `templates/foo.yaml` files support inside Helm charts so that we can generate value/secret strings which can use templating to compose things from smaller secret values (for example, creating a maven `settings.xml` file or docker `config.json` which includes many user/passwords for different registries).
```

content/en/docs/install-setup/installing/boot/ingress.md
```diff
@@ -0,0 +1,85 @@
+---
+title: Ingress
+linktitle: Ingress
+description: Configure ingress
+weight: 65
+---
+
+## Ingress
+
+If you don't specify anything in your [jx-requirements.yml](https://github.com/jenkins-x/jenkins-x-boot-config/blob/master/jx-requirements.yml) file then boot will default to using HTTP (rather than HTTPS) and using [nip.io](https://nip.io/) as the DNS mechanism.
```
jha-cloudbees, line 10:
```suggestion
If you don't specify anything in your [jx-requirements.yml](https://github.com/jenkins-x/jenkins-x-boot-config/blob/master/jx-requirements.yml) file then Boot will default to using HTTP (rather than HTTPS) and using [nip.io](https://nip.io/) as the DNS mechanism.
```

content/en/docs/install-setup/installing/boot/ingress.md
```diff
@@ -0,0 +1,85 @@
...
+
+After running boot your `jx-requirements.yml` may look like:
```
jha-cloudbees, line 12:
```suggestion
After running Boot your `jx-requirements.yml` may look like:
```

hferentschik:
> I think you are assuming too much context on the user's part, what may be technically correct is not necessarily what the mental model of a new user looking for information would be.

Agreed. However, the mental model must still be consistent. You cannot describe something on one level with terms and definitions which are not explained yet. Apart from the fact that the information on this page, for example, is incomplete and wrong, it needs the concept of Boot first to make sense.

 > We need to meet a user's mental model and then map it to our concepts, not the inverse.

I think I do.

> And on the sidebar, you want actions, not functions.

- Create cluster
- Install jx
- Run Boot

sounds actionable to me. The headings underneath are maybe more "functional", but there is a tradeoff between too much repetition and readability. I think one does not need to repeat "configure" all the time. But I can add it if you think it helps.

> download tends to mean "getting access to the thing"

-1 that's imo not all what it means. Downloading imo opinion has a very specific meaning almost solely limited to "downloading" a given binary. That's imo something completely different than "getting access to something". The latter could mean much more than the actual download. It could involve getting a license or a login or configuring a firewall to open up a port to "accessing" the functionality I am after.

> Renaming this to "Installing Jenkins X":

Running Boot is much more than "Installing Jenkins X". Why would you try to "hide" that? You need to understand the concept of Boot to be successful with Jenkins X. IMO it needs to be described early and well.

content/en/docs/install-setup/installing/boot/secrets.md
```diff
@@ -0,0 +1,165 @@
...
+
+The Vault configuration options are in this case generated and should not be modified.
+They also differ between cloud providers.
+
+#### External
+
+In the case where you have an existing Vault instance and you want Jenkins X to store its secrets there, you also set `secretStorage: vault`.
+On top of this you have to specify the mandatory _url_ and _serviceAccount_ options.
+
+_url_ specifies the URL to your existing Vault instance and _serviceAccount_ is the Kubernets service account which is allowed to authenticate using Vault's [Kubernetes Auth Method](https://www.vaultproject.io/docs/auth/kubernetes).
```
jha-cloudbees, line 75:
```suggestion
_url_ specifies the URL to your existing Vault instance and _serviceAccount_ is the Kubernetes service account which is allowed to authenticate using Vault's [Kubernetes Auth Method](https://www.vaultproject.io/docs/auth/kubernetes).
```

content/en/docs/reference/components/vault.md
```diff
@@ -4,262 +4,98 @@ linktitle: Vault
...
+```
 
-6.  The program will ask for an `URLTemplate`. Press `Enter` to
-use the default value.
+## Rotating secrets
 
-7.  The program will verify your answers to the previous questions in
-summary and prompt you to approve the Vault creation (default is
-`Yes`).
+To rotate a secret you follow the steps described in the [Updating a secret](/docs/reference/components/vault#updating-secrets) section followed by rerunning [Boot](/docs/install-setup/installing/boot) (`jx boot`).
```
jha-cloudbees, line 99:
```suggestion
To rotate a secret, follow the steps described in the [Updating a secret](/docs/reference/components/vault#updating-secrets) section followed by rerunning [Boot](/docs/install-setup/installing/boot) (`jx boot`).
```

hferentschik:
> Move secrets down so that it's part of the grouping with the others

not sure what you mean? It is part of the group with the others, right? It just happens to be the first in the group, since it happens to be the first thing one needs to sort out. But if you tell me where exactly you want to have it,  I might better understand what you are after.

hferentschik:
> Renaming this to "Download Jenkins X"

You are not "Downloading Jenkins X", you are downloading `jx`. You are downloading the installer so to say, not the actual thing. I think it is essential that we start nailing the terminology. A lot of confusion stems atm from the fact that we don't have a clear terminology.

ethangj:
"You are not "Downloading Jenkins X", you are downloading jx."

This is an important place I strongly disagree, as a new user you can not expect me to care about these details. I care about the outcome - I want Jenkins X up and running, and you need to tell me what to do in what order - download, install, configure. We have to think of this as product documentation, not technical specifications. At the **end** of following the docs, the user should know the difference between JX and Jenkins X, but they shouldn't need to know to get started.

"not sure what you mean? It is part of the group with the others, right?"

It's cut off by "How It Works", which feels weird. It's conceptually closer to the things below "How it Works" then it is the ones above it.

"Run boot" is just not empathetic to a user, a user can't be expected to know what boot is in order to know where to go in the navigation to proceed. Boot is our word, not the customers. If you did a live session with a customer, asking them to narrate what they wanted to do, nobody would say "my goal with Jenkins x is to use JX boot quickly", they'd say it was to install JX quickly.

hferentschik:
> "Run boot" is just not empathetic to a user, a user can't be expected to know what boot is in order to know where to go in the navigation to proceed.

I don't expect him to know. I am going to explain it. "Install and Setup" provides the context and atm there is even the more detailed "Install Jenkins X". Under there we have necessary various sub-tasks. At no point, this implies that you know what "Boot means".

 > Boot is our word, not the customers. If you did a live session with a customer, asking them to narrate what they wanted to do, nobody would say "my goal with Jenkins x is to use JX boot quickly", they'd say it was to install JX quickly.

Agreed and to do that you would need to explain some concepts. If you were to explain to someone how to drive a car, you would not just sit next to him and tell him to start driving, right? Even though that's what he wants. You would first understand the basic parts involved, aka steering wheel, gear shift, etc.

hferentschik:
> It's cut off by "How It Works", which feels weird. It's conceptually closer to the things below "How it Works" then it is the ones above it.

How can it be cut off by "How it works"? "How it works" is not in-between "Secrets" and "Git Providers" and the others. It's above. It's the first in this "group", even though the group in this context is imo just odd. On this one, I really don't get what you are after or what you are suggesting to get a different grouping.

ethangj:
For context, we met offline and agreed to push this forwarding in the short term so as not to be a blocker and we will take up larger refactoring of some fo this content and structure sometime soon.

hferentschik:
@jha-cloudbees - I applied your typo fixes. Thanks
@ethangj - "installing jx" is now "downloading jx"

PR #2857, opened 2020-06-09
chore: Add capabilities matrix

deanesmith:
High level guidance across Kubernetes providers of what key jx features work and what do not

Signed-off-by: deanesmith <dsmith@cloudbees.com>

### DIFF
--- a/content/en/about/capabilities.md
+++ b/content/en/about/capabilities.md
@@ -1,0 +1,27 @@
+---
+title: Jenkins X Capabilities Matrix
+linktitle: Capabilities Matrix
+description: Matrix that describes the current state of Jenkins X capabilities for major Kubernetes platforms
+type: docs
+weight: 20
+---
+
+## Jenkins X Capabilities Matrix
+
+| Capability | GKE | EKS | OpenShift 3.11 |
+| --- | :---:| :---: | :---: |
+| **Terraform scripts for cluster creation** | Yes | Yes | No |
+| **`jx boot` for installation** | Yes | Yes | Yes |
+| **Vault for secret storage** | Yes | Yes | Self Provisioned |
+| **Kubernetes Versions** | | | |
+| - 1.13 | Yes | No | N/A |
+| - 1.14 | Yes | Yes | N/A |
+| - 1.15 | Preview | Preview | N/A |
+| - 1.16 | Preview | Preview |  N/A |
+| **Source Control Providers** | | | |
+| - Github | Yes | Yes | Yes |
+| - Github Enterprise | Yes | Yes | Yes |
+| - Gitlab CE | Preview | Preview | Preview |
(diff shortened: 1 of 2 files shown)


ankitm123:
What about the case where vault works in EKS 1.15, and does not work in EKS 1.16 (the vault operator chart had to be updated for it to be able to boot) - how would that be represented? Vault is just one example, there might be more ...

content/en/about/capabilities.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+## Jenkins X Capabilities Matrix
+
+| Capability | GKE | EKS | AKS | OpenShift 3.11 |
+| --- | :---:| :---: | :---: | :---: |
+| **Terraform scripts for cluster creation** | Yes | Yes | No | No |
+| **`jx boot` for installation** | Yes | Yes | No | Yes |
+| **Vault for secret storage** | Yes | Yes | No | Self Provisioned |
+| **Nexus** | Yes | Yes | No | Yes |
```
hferentschik, line 16: Why is Nexus worth mentioning? That seems to me a bit random and too specific on this level of abstractio.

hferentschik:
> What about the case where vault works in EKS 1.15, and does not work in EKS 1.16 (the vault operator chart had to be updated for it to be able to boot) - how would that be represented? Vault is just one example, there might be more ...

I feel the same way. The format of this table is in itself limiting, but I guess there is no easy way to depict the complexity of things. It might be better than nothing.

> Things can be version specific and platform specific at the same time, and that's where we will have issues with in this representation.

+1. Also, things might not be Kubernetes (cloud provider) specific at all like the SCM providers. 

I am wondering whether AKS should be even in the table, because we effectively have no AKS.

content/en/docs/install-setup/installing/boot/clouds/_index.md
```diff
@@ -19,24 +19,6 @@ However if you try doing any significant work with Kubernetes particularly with
 
 This section of the documentation will try guide you through the best way to use Jenkins X on different cloud providers.
 
-## Cloud Provider Support Maturity
+Refer to the [Jenkins X Capabilities Matrix](/about/capabilities) for cloud provider support maturity.
```
hferentschik, line 22: 👍

content/en/about/capabilities.md
```diff
@@ -0,0 +1,27 @@
...
+title: Jenkins X Capabilities Matrix
+linktitle: Capabilities Matrix
+description: Matrix that describes the current state of Jenkins X capabilities for major Kubernetes platforms
+type: docs
+weight: 20
+---
+
+## Jenkins X Capabilities Matrix
+
+| Capability | GKE | EKS | OpenShift 3.11 |
```
hferentschik, line 11: 👍

PR #2887, opened 2020-06-17
docs: updating the git provider support documentation

### LINKED ISSUE #2884: Duplicate documentation on how to configure your git provider
At the moment we have:

- http://jenkins-x.io/docs/install-setup/configure-scm/ (configure-scm.md)
- https://jenkins-x.io/docs/install-setup/installing/boot/#git (as part of _index.md of the boot directory)

The former is incomplete and refers to changing the requirement files even though it has not been introduced yet. The concept of Boot and `jx-requirements.yml` needs to be first introduced before one can start talking about changing any settings in this file.

Also, just changing the setting as suggested in http://jenkins-x.io/docs/install-setup/configure-scm/ is not going to work unless the Boot pipeline is triggered. 

Ideally, the change is not made and applied locally, but via a pull request. 

For that reason, I suggest to delete /docs/install-setup/configure-scm/ for now and focus on getting the documentation as part of Boot correct. Whether eventually, we want to change the navigation or context to make things "easier" is imo a follow-up issue. The primary concern for this issue is to remove duplicated and false documentation.

hferentschik:
- removing incomplete and duplicated content/en/docs/install-setup/configure-scm.md
- updated content/en/docs/install-setup/installing/boot/git-provider.md

fixes #2884


hferentschik:
@abayer - any idea what the _gitName_ property is good for? I guess you can name your provider, but where does that do you any good? Where would you display that or do a provider lookup by name?

Second question, what's the deal with the git provider configuration per environment in `EnvironmentConfig`? We keep talking about changing the git provider on a global `ClusterConfig` level, but at least as per schema, you can do this per environment as well. Does that work? And what are the effects in this case?

hferentschik:
Turns out I found part of the answer in this page - https://jenkins-x.io/docs/reference/components/git/. Still not sure what actually works and whatnot. I would, for example, think that the `jx create git ...` commands won't work. At least not in conjunction with Boot.

abayer:
@hferentschik Honestly, I have no clue what `gitName` is for. Same w.r.t. the per-environment stuff.

hferentschik:
> Honestly, I have no clue what gitName is for. Same w.r.t. the per-environment stuff.

That's what I feared. Looking at the code, I don't think it is really used. Also as it turns out `jx create git server` does persist into the auth service AND create a GitService CRD. However, `jx get git` is only looking at what's in the auth service. 

https://github.com/jenkins-x/jx/blob/cff0301241c853ce8d68bddb3009f6585d39ef97/pkg/cmd/create/create_git_server.go#L126

https://github.com/jenkins-x/jx/blob/cff0301241c853ce8d68bddb3009f6585d39ef97/pkg/cmd/get/get_git.go#L54

So I would think most of what's here - https://jenkins-x.io/docs/reference/components/git/ - is obsolete or half complete. Git servers per environment I would suggest to not even mention for now, since it is unclear to me what it would even do. 

This really needs a plan on how this should be taken forward. Might be worth cherry-picking the morsels of information which make sense from https://jenkins-x.io/docs/reference/components/git/. and otherwise delete it.

ethangj:
Looks fine to me, but someone technical should approve.

hferentschik:
> Looks fine to me, but someone technical should approve.

no worries. I'll make some more changes and then I let for example @abayer have another look at it.

content/en/docs/install-setup/installing/boot/git-provider.md
```diff
@@ -6,152 +6,85 @@ keywords: [git]
...
+  gitKind: github
 ```
 
-### GitHub Enterprise
+## GitHub Enterprise
+
+[GitHub Enterprise](https://github.com/enterprise) supports the same features as the github.com service but scaled for on-premises deployment on local networks.
 
-The configuration is similar to the above but you need to specify the URL of the `gitServer` (if it differs from https://github.com) and `gitKind: github`
+The configuration is similar to the above but you need to specify the URL of the `gitServer` and `gitKind: github`:
```
jha-cloudbees, line 39:
```suggestion
The configuration is similar to the above but you need to specify the URL of the `gitServer` and `gitKind: github`. For example:
```

PR #2914, opened 2020-06-23
fix: changing git to SCM

ethangj:
Changing Git to SCM because it's what more people are likely to search for and parse the page for.

### DIFF
--- a/content/en/docs/install-setup/installing/boot/git-provider.md
+++ b/content/en/docs/install-setup/installing/boot/git-provider.md
@@ -1,5 +1,5 @@
 ---
-title: Configure SCM / Git
+title: Configure SCM Providers / Git Providers
 linktitle: Configure SCM Providers
 description: Configure SCM Providers / Git providers supported by Jenkins X
 keywords: [git]
--- a/content/en/docs/install-setup/installing/boot/git-provider.md
+++ b/content/en/docs/install-setup/installing/boot/git-provider.md
@@ -1,9 +1,11 @@
 ---
-title: Configure Git Provider
-linktitle: Configure Git Provider
-description: Configure Git providers supported by Jenkins X
+title: Configure SCM / Git
+linktitle: Configure SCM Providers
+description: Configure SCM Providers / Git providers supported by Jenkins X
 keywords: [git]
 weight: 40
+aliases:
+  - /docs/install-setup/configure-scm/
 ---
 
 Jenkins X supports several different Git providers via webhooks (user-defined HTTP callbacks).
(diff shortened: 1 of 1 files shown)


content/en/docs/install-setup/installing/boot/git-provider.md
```diff
@@ -1,9 +1,11 @@
 ---
-title: Configure Git Provider
-linktitle: Configure Git Provider
-description: Configure Git providers supported by Jenkins X
+title: Configure SCM / Git
```
hferentschik, line 2: Why don't you want to use Git Providers? You are not configuring Git really here.

ethangj:
I think SCM is more likely to be what people scan for in this context. The way I see it, Git is the technology, the SCM is the tool. I know there's another view where SCM is Mercurial vs Git, but I think that's a less common (even if it's technically correct.) 

People will say "my SCM is Bitbucket", not "my Git is Bitbucket."

hferentschik:
> I think SCM is more likely to be what people scan for in this context. 

Why not leverage tags for that? So SCM could still be associated with this page and the tag could even display somewhere, but we can use the correct names. 

> The way I see it, Git is the technology, the SCM is the tool. 

I guess you mean the other way around, right? And one way or the other the only SCM we support is Git. What differs is the provider. We need to have a concise terminology/glossary otherwise we will never be able to talk w/o misunderstanding about things. Whether there are other related "search terms" associated to the concept is imo a different discussion.

> I know there's another view where SCM is Mercurial vs Git, but I think that's a less common (even if it's technically correct.)

I think any developer will understand "Git Provider"  as a term. It triggers the right connections. SCM is and will be ambiguous.
Look here for example - https://codefresh.io/docs/docs/integrations/git-providers/

content/en/docs/install-setup/installing/boot/git-provider.md
abayer: My personal preference would be for `Configure SCM Provider` - longstanding term-of-art for GitHub et al. But the "Provider" bit is, IMO, needed.

hferentschik: > Configure SCM Provider

I could live with that. It is more generic and potentially we support another SCM at some stage (conceptually).

> But the "Provider" bit is, IMO, needed.

+1, this is my point as well.

hferentschik: @ethangj - should the title not change as well?

ethangj:
@hferentschik updated

PR #2925, opened 2020-06-25
docs: updating Terraform module usage for GKE and EKS

hferentschik:
fixes #2023

### DIFF
--- a/content/en/docs/install-setup/create-cluster/eks.md
+++ b/content/en/docs/install-setup/create-cluster/eks.md
@@ -21,7 +21,7 @@ The source is in the [terraform-aws-eks-jx](https://github.com/jenkins-x/terrafo
 The EKS Terraform module for Jenkins X allows you to create an [EKS](https://aws.amazon.com/eks/) cluster for installation of Jenkins X.
 You need the following binaries locally installed and configured on your _PATH_:
 
-- `terraform` (=> 0.12.17)
+- `terraform` (~> 0.12.17)
 - `kubectl` (>=1.10)
 - `aws-iam-authenticator`
 - `wget`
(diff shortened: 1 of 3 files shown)


content/en/docs/install-setup/create-cluster/eks.md
```diff
@@ -21,7 +21,7 @@ The source is in the [terraform-aws-eks-jx](https://github.com/jenkins-x/terrafo
 The EKS Terraform module for Jenkins X allows you to create an [EKS](https://aws.amazon.com/eks/) cluster for installation of Jenkins X.
 You need the following binaries locally installed and configured on your _PATH_:
 
-- `terraform` (=> 0.12.17)
+- `terraform` (~> 0.12.0)
```
ankitm123, line 24: This is something I always forget to fix in the readme of the module, but it needs to be =>0.12.17. Now that 0.13 beta is out, we may want to do `~> 0.12.17`. See: https://github.com/jenkins-x/terraform-aws-eks-jx/issues/47#issuecomment-626251755

hferentschik: yeah, should be corrected. I keep forgetting as well. Probably not so important after all, but it's just nice to be accurate. I have an outstanding PR on the Terraform module. I can do it there, so the next time we sync it will be right. I made some other changes to the README there which make the sync easier, but there are will always be some customization needed (heading levels, links, some sentences which need to change depending of the location of the docs). But imo it is not too bad anymore.

hferentschik: I updated https://github.com/jenkins-x/terraform-aws-eks-jx/pull/98 and will push an update here as well ...

hferentschik: pushed an update

PR #2957, opened 2020-07-13
fix: remove highlightjs and use chroma

### LINKED ISSUE #2943: Fix syntax highlighting for terraform code blocks
Syntax highlighting for terraform codeblocks is broken in the website.
This is what is rendered for any code block which starts with ` ```terraform `
image
Removing ` ```terraform ` renders the code block properly.
image

Two issues:
* jx-docs use highlight.js internally for synatx [highlighting](https://jenkins-x.io/community/documentation/reference/#code-examples), while hugo uses chroma. For languages supported by highlight.js, it adds a class `hljs <language-name>`, for example for bash, it adds `hljs nginx` (not sure why nginx), but does not add any `hljs` class for terraform. We should remove highlight.js, and use hugo's chroma syntax highlighting (Not sure, why we use both). 

* Hugo uses chroma for syntax highlighting, which has a bug for terraform syntax highlighting which was fixed in a later [release](https://github.com/alecthomas/chroma/releases/tag/v0.7.2). 
The extended hugo docker image used by the doc site uses [0.65.3](https://github.com/jenkins-x/hugo-extended-docker/blob/master/HUGO_VERSION), which uses [0.7.1](https://github.com/gohugoio/hugo/blob/v0.65.3/go.mod#L8) version of chroma.

We should remove highlight.js and update the hugo version (by using the [recommended](https://gohugo.io/getting-started/installing/#docker) docker image for hugo) to fix this issue. The current size of the docker image for hugo extended is 1.93 GB (the extended version of the recommended one is around 500 MB).

ankitm123:

fixes #2943

### DIFF
--- a/assets/scss/_code.scss
+++ b/assets/scss/_code.scss
@@ -11,13 +11,13 @@
         border: none;
         background-color: transparent;
 
-        pre, div {
+        div {
             background-color: transparent !important;
         }
 
         pre {
             margin: 0;
-            padding: 0;
+            padding: $pre-padding;
         }
     }
 
@@ -45,12 +45,19 @@
      
 
         > code {
-            padding: 15px;
             margin: 0;
             font-size: 100%;
             word-break: normal;
             white-space: pre;
             border: 0;
         }
+    }
+}
+
+//
+.highlight {
+    pre {
+        margin: 0;
+        padding: $pre-padding;
     }
 }
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,8 +1,8 @@
 version: "2"
 services:
   server:
-    image: gcr.io/jenkinsxio/hugo-extended:0.65.3-6
-    command: ["hugo", "server", "-s", "/src", "-D", "--bind", "0.0.0.0", "--ignoreCache"]
+    image: gcr.io/jenkinsxio/hugo-extended:0.74.0-10
+    command: ["server", "-s", "/src", "-D", "--bind", "0.0.0.0", "--ignoreCache"]
     volumes:
       - .:/src
     ports:
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-HUGO_VERSION := v0.65.3
+HUGO_VERSION := v0.74.0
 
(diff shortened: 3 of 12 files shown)


assets/scss/_code.scss
```diff
@@ -11,7 +11,11 @@
         border: none;
         background-color: transparent;
 
-        pre, div {
+        //pre, div {
```
ankitm123, line 14: The pre block was changing the default background from dark to transparent

ankitm123:
The only issue I see is that, I need to have a blank line between the backtick and the code snippet, else the indentation is a bit messed up.

docker-compose.yml
```diff
@@ -1,8 +1,8 @@
 version: "2"
 services:
   server:
-    image: gcr.io/jenkinsxio/hugo-extended:0.65.3-6
-    command: ["hugo", "server", "-s", "/src", "-D", "--bind", "0.0.0.0", "--ignoreCache"]
+    image: klakegg/hugo:0.73.0-ext-alpine
```
ankitm123, line 4: This is just for preview/demo, need to discuss before removing the hugo extended base image

assets/scss/_code.scss
```diff
@@ -45,7 +45,6 @@
      
 
         > code {
-            padding: 15px;
```
ankitm123, line 48: This was giving the extra indentation to the first line in a code snippet.

MarckK:
Thank you, @ankitm123! Looks good =) There is a bit of extra dark highlighting going on that would be great to remove, eg:
Screenshot 2020-07-13 at 12 09 26

ankitm123:
> Thank you, @ankitm123! Looks good =) There is a bit of extra dark highlighting going on that would be great to remove, eg:

This seems like a chroma/hugo issue. Let me check if there's a fix.

MarckK:
Awesome! Thank you, @ankitm123

ankitm123:
/hold

Getting compilation issues with the latest hugo version, checking ...

MarckK:
Looks fantastic! Thank you, @ankitm123 :)

/lgtm

PR #3007, opened 2020-08-04
docs: update FAQ on secrets management

MarckK:
Signed-off-by: Kara de la Marck <MarckK@users.noreply.github.com>

### DIFF
--- a/content/en/docs/resources/faq/using/_index.md
+++ b/content/en/docs/resources/faq/using/_index.md
@@ -53,9 +53,9 @@ You may need to modify your helm charts to add extra helm configuration if the c
 
 ## How do I manage secrets in each environment?
 
-We’re using sealed secrets ourselves to manage our production Jenkins X install for all of our CI/CD - so the secrets get encrypted and checked into the git repo of each environment. We use the [helm-secrets](https://github.com/futuresimple/helm-secrets) plugin to do this.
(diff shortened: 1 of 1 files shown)


content/en/docs/resources/faq/using/_index.md
```diff
@@ -53,9 +53,9 @@ You may need to modify your helm charts to add extra helm configuration if the c
 
 ## How do I manage secrets in each environment?
 
-We’re using sealed secrets ourselves to manage our production Jenkins X install for all of our CI/CD - so the secrets get encrypted and checked into the git repo of each environment. We use the [helm-secrets](https://github.com/futuresimple/helm-secrets) plugin to do this.
+[Hashicorp Vault](https://www.vaultproject.io/) is the preferred way in Jenkins X to manage secrets. For example, the GitHub personal access token generated for the pipeline bot is stored in Vault. [Read more about using Vault to manage your secrets with Jenkins X](https://jenkins-x.io/docs/reference/components/vault/).
```
ankitm123, line 56: Actually we use bank-vault or the vault operator.

MarckK: Yeah, it's in the Vault docs that are linked to: 

"There are two ways you can use Vault in Jenkins X. If you are already a Vault instance, you can configure Jenkins X to use this instance to store its secrets. This way you have a central location for managing all secrets of your infrastucture. If you do not have a Vault instance, Jenkins X can, depending on where you run your cluster, install Vault as part of the installation process. In this case, the Banzai Cloud Bank-Vaults operator is provisioned in the Jenkins X development namespace."

We shouldn't repeat paragraphs of documentation in multiple places. It's better to link to one main source of documentation for an issue or topic.

PR #3020, opened 2020-08-16
docs: add how to get the new jx3 alpha CLI commands

rawlingsj:


ankitm123:
Is the change to docsy submodule intended? Probably time, we update docsy hmmm ...

PR #3022, opened 2020-08-17
docs: tiny edits

MarckK:


MarckK:
Double checking that we do demand that contributors to the docs sign DCO.  ;)

MarckK:
@ankitm123 can you approve this for me? I want to see if it merges with no problem, even without the commit signed.

ankitm123:
We dont have the requirement for signing commits yet, so pretty sure, it will pass. However, this should be implemented :)

PR #3033, opened 2020-08-23
fix: fix docs navbar to top

### LINKED ISSUE #3015: Fixing the navbar of jx-docs site.
This issue is a dependency on [#2877](https://github.com/jenkins-x/jx-docs/issues/2877).

Primarily while adding the dark-mode switcher on the navbar, it would be beneficial to the user to fix the navbar to the top of the page so as to access the switcher and other tabs without the need to scroll on the top.

Feel free to share your views on this issue.

ysahil97:
fixes #3015

### DIFF
--- a/assets/scss/_sidebar-tree.scss
+++ b/assets/scss/_sidebar-tree.scss
@@ -27,19 +27,19 @@
         ul {
             padding: 0;
             margin: 0;
- 
+
         }
 
         @include media-breakpoint-up(md) {
             & > ul {
                 padding-left: .5rem;
-                
+
             }
         }
 
 
         padding-left: 0;
-   
+
     }
 
     &__section-title {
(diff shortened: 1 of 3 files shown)


ankitm123:
probably meant #3015? @ysahil97

ankitm123:
@ysahil97 looks great :100: 

The only issue is that the upper portion of the sidebar gets hidden, when scrolling down the page.
image

ysahil97:
I have investigated this for a while, we can try to fix the sidebar, leaving only main content to scroll. The only issue that is coming up with this is that the right border of the sidebar would not run till the end of the page, causing background color discontinuity. A suggestion could be to make the sidebar background color white, but I am open to other suggestions as well.

image

ankitm123:
I am fine with making the background color white :+1:

PR #3035, opened 2020-08-24
fix: tell minikube to use VM

polothy:
When trying to use Docker with minikube, I got this notification after minikube create the cluster:

```
Due to networking limitations of driver docker on darwin, ingress addon is not supported.
Alternatively to use this addon you can use a vm-based driver:

	'minikube start --vm=true'

To track the update on this work in progress feature please check:
https://github.com/kubernetes/minikube/issues/7332
```


ankitm123:
@polothy need to change commit message and PR title (they are failing the checks).

Issue #3037, opened 2020-08-25
Issue Templates for Jenkins X docs repo


borntorock:
Need to create the Issue Templates for the users reference and usage while submitting the issue on the Jenkins X docs repo.

PR #3038, opened 2020-08-25
fix: creation of Issue Templates for users reference

borntorock:
Fixes https://github.com/jenkins-x/jx-docs/issues/3037

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,26 @@
+---
+name: Bug Report
+about: Report a bug encountered with Jenkins X docs
+labels:
+- kind/bug
+---
+**This is a Bug Report**
+
+<!--Required Information-->
+**Problem:**
+
+
+**Proposed Solution:**
+
+
+**Page to Update:**
+https://jenkins-x.io/docs/....
+
+
+**Steps to Reproduce:**
+
+
+<!--Optional Information (remove the comment tags around information you would like to include)-->
+<!--Jenkins-X Version:-->
+
+<!--Additional Information:-->
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,5 @@
+blank_issues_enabled: true
+contact_links:
+  - name: Ask a question
+    url: https://jenkinsx.discourse.group/
+    about: For increased visibility, please post questions on the discourse forum.
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -1,0 +1,17 @@
+---
+name: Feature Request
+about: Suggest a/an feature/enhancement to the Jenkins X docs project
+labels:
+- kind/feature
+---
+**This is a Feature Request**
+
+
+**What would you like to be added**
+<!-- Describe as precisely as possible how this feature/enhancement should work from the user perspective. What should be changed, etc. -->
+
+**Why is this needed**
+
+**Comments**
+<!-- Any additional related comments that might help. Drawings/mockups would be extremely helpful (if required). -->
+


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,23 @@
+---
+name: Bug Report
+about: Report a bug encountered with Jenkins X docs
+labels:
+- area/docs
+- kind/bug
```
borntorock, line 6: Need to see if these labels for the bug report issue are correct.

.github/ISSUE_TEMPLATE/config.yml
```diff
@@ -0,0 +1,5 @@
+blank_issues_enabled: false
```
borntorock, line 1: Do we still need to have the possibility of creating blank issues for the users in jx-docs repo?

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -0,0 +1,18 @@
+---
+name: Feature Request
+about: Suggest a/an feature/enhancement to the Jenkins X docs project
+labels:
+- area/docs
+- documentation
```
borntorock, line 6: Are these labels appropriate or should we include `kind/feature` or any other label?

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,23 @@
+---
```
borntorock, line 1: Apart from Bug Report template and feature request, do we need to also have security policy in this repo as present in jx repo?

ankitm123:
Thinking about it a bit more, since these are docs, I think just a blank template, and a link to discourse (for asking questions - use the yml for it) should be enough imo. wdyt? @borntorock @MarckK

MarckK:
That sounds @ankitm123. 

Linking to discourse for any content issues (which are likely to be most of the docs issues) and issues here for bugs with the docs site or suggested improvements to its structure, etc.

borntorock:
I have created these templates with reference to the Kubernetes website / docs repo. See [here](https://github.com/kubernetes/website/issues/new/choose). 

A blank template would be similar to what we have currrently. IMHO, I think a template should be available for the users (for reference) while opening up a bug for docs site. Though following the template strictly is not a compulsion for the users.

.github/ISSUE_TEMPLATE/config.yml
ankitm123: Only blank is what we should have imo. Anything more sounds excessive to me honestly.

borntorock: @MarckK What do you think? Let me know, so as to finalize this PR.

ankitm123:
If you see some of the issues in the kubernetes website, I really dont see the value in those templates, as people just ignore them :smile: The most recent ones are the one to do with translation, which we are not mature or big enough to implement imo. 

However, I am ok with making it exactly like the kubernetes doc repo issue section. May be we can see what the user does, and if ppl follow it, stick to it, if not, then remove it. wdyt?

Also, you need to change the commit message and PR title format.

borntorock:
@ankitm123 Sounds good. But  I don't think people will actually and always follow the template. But this is just a medium to gather as much information as we can regarding the bug or feature request they are making. 

On the other hand, I understand your concern and since these templates doesn't enforce the users to follow the template strictly, they can definitely overwrite and proceed with the creation of blank issue.

borntorock:
@ankitm123 Could you Please provide appropriate example for this `Also, you need to change the commit message and PR title format.` I'm not sure why the semantic checks are failing. I tried changing the commit message as well as PR title still some checks are not successfull. Thanks!

.github/ISSUE_TEMPLATE/config.yml
ankitm123: We should have blank issues.

.github/ISSUE_TEMPLATE/feature_request.md
ankitm123: `area/docs` and `documentation` labels in a docs repo dont add much information.
I prefer `kind/feature` (or `kind/enhancement`) and `kind/bug`.

.github/ISSUE_TEMPLATE/bug_report.md
ankitm123: Remove the area/docs, not required imo. Every issue opened here would be area/docs.

ankitm123:
@borntorock I am fine with the extra feature and bug templates. Left a few comments for you to address.

.github/ISSUE_TEMPLATE/feature_request.md
borntorock: you mean `kind/bug` for both bug report and feature request?

borntorock:
just addressed your comments and changed accordingly. Let me know, if this looks good to  be merged now.

PR #3042, opened 2020-08-28
preview environment creation details are incorrect

troyhart:
I guess I could be wrong, but this is what I understand.


MarckK:
Thank you for your PR, @troyhart!

/lgtm

PR #3052, opened 2020-09-08
docs: add easier dco signoff

MarckK:


content/en/community/code/_index.md
```diff
@@ -321,48 +321,66 @@ by viewing the `Checks` tab in the GitHub pull request.
...
-    #!/bin/sh
+1. The easiest way is to add the `--signoff` or `-s` flag when creating your commit. This flag must be added to each commit you would like to sign.
 
-    COMMIT_MSG_FILE=$1  # The git commit file.
-    COMMIT_SOURCE=$2    # The current commit message.
+```sh
+git commit -m "docs: my commit message" -s
+```
+
+2.  You can use a hook to make sure all your commits messages are signed off. Run:
```
ankitm123, line 334: This should be the first option, as option 1 is closely related to option 3.

content/en/community/code/_index.md
```diff
@@ -321,48 +321,66 @@ by viewing the `Checks` tab in the GitHub pull request.
...
+```sh
+mkdir -p ~/.git-templates/hooks
+```
+```sh
+git config --global init.templatedir ~/.git-templates
+```
+
+Then add this to `~/.git-templates/hooks/prepare-commit-msg`:
```
ankitm123, line 342: The indentation seems off in the preview.

content/en/community/code/_index.md
MarckK: I like keeping option 1 as the first as it is, by far, the simplest for someone new to the project. And we want to make the onboarding of new contributors as simple as possible (there's more work to be done there).

You are right that 1 and 3 are closely related, so maybe we move option 3 up to being the second listed option. Makes sense as we already are recommending it among the options.

So then we would have: 1) simplest/fastest way to sign (just add a flag to commit message) 2) Our recommended way to set up commit signing 2) and an alternative way, for those who want one.

ankitm123: iirc, u still need to set up the gpg key to sign the commits, so even adding the s flag, would need some set up. So imo, 2 is the easiest way to sign (sort of) the commits. Reference: https://docs.github.com/en/github/authenticating-to-github/signing-commits

MarckK: Edited PR to differentiate between `git commit -s, --signoff` and `git commit -S, --gpg-sign`.

PR #3057, opened 2020-09-15
feat: prism support for syntax highlighting and copy button on code blocks

### LINKED ISSUE #3049: add a copy button to code examples
**This is a Feature Request**

**What would you like to be added**

Code examples in tutorials, READme files, and guides could use a copy button so people don't miss a single thing when they copy code.

**Why is this needed**
Better user experience.

**Comments**

Once this PR, https://github.com/google/docsy/pull/325, is merged Jenkins X docs should be able to add copy button to code examples quite easily. 🤞

borntorock:
Resolves #3049 

Signed-off-by: borntorock <borntorock@users.noreply.github.com>

### DIFF
--- a/layouts/partials/head-css.html
+++ b/layouts/partials/head-css.html
@@ -12,3 +12,7 @@
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
 <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.15.10/styles/monokai.min.css">
 <link rel="stylesheet" href='{{ "css/style.css" | absURL }}' />
+{{ with .Site.Params.prism_syntax_highlighting }}
+<!-- stylesheet for Prism -->
+<link rel="stylesheet" href='{{ "css/prism.css" | absURL }}' />
+{{ end }}
--- a/layouts/partials/scripts.html
+++ b/layouts/partials/scripts.html
@@ -83,4 +83,8 @@
 	}
 
 </script>
+{{ with .Site.Params.prism_syntax_highlighting }}
+<!-- scripts for prism -->
+<script src='/js/prism.js'></script>
+{{ end }}
 {{ partial "hooks/body-end.html" . }}
--- a/config.toml
+++ b/config.toml
@@ -126,6 +126,9 @@ github_project_repo = "https://github.com/jenkins-x/jx"
 # Google Custom Search Engine ID. Remove or comment out to disable search.
 gcs_engine_id = "011737558837375720776:fsdu1nryfng"
 
+# Enable syntax highlighting and copy buttons on code blocks with Prism
+prism_syntax_highlighting = true
+
 # User interface configuration
 [params.ui]
 # Enable to show the side bar menu in its compact state.
--- a/content/en/docs/_index.html
+++ b/content/en/docs/_index.html
@@ -53,7 +53,7 @@ layout="home"
           1. Create a <code>main.tf</code> file in an empty directory and add the code snippet below.
           You will need to provide the GCP project id in which to install the cluster.
         </p>
-{{<highlight javascript>}}
+{{<highlight terraform>}}
 module "jx" {
   source = "jenkins-x/jx/google"
 
@@ -89,7 +89,7 @@ terraform apply
         <p>
           1. Create a <code>main.tf</code> file in an empty directory and add the code snippet below.
         </p>
-{{<highlight javascript>}}
+{{<highlight terraform>}}
 module "eks-jx" {
   source  = "jenkins-x/eks-jx/aws"
 }
--- a/static/css/prism.css
+++ b/static/css/prism.css
@@ -1,0 +1,189 @@
+/* PrismJS 1.21.0
+https://prismjs.com/download.html#themes=prism-tomorrow&languages=markup+css+clike+javascript+bash+css-extras+docker+go+groovy+makefile+yaml&plugins=toolbar+copy-to-clipboard */
+/**
+ * prism.js tomorrow night eighties for JavaScript, CoffeeScript, CSS and HTML
+ * Based on https://github.com/chriskempson/tomorrow-theme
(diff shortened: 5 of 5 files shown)


ankitm123:
I think you also need to pull the latest docsy master (included as a submodule here). I dont see any copy to clipboard in the preview. Also most probably by default terraform syntax highlighting may not be enabled, so u have to add the required css/js files.

borntorock:
> I think you also need to pull the latest docsy master (included as a submodule here). I dont see any copy to clipboard in the preview. Also most probably by default terraform syntax highlighting may not be enabled, so u have to add the required css/js files.

Yes, this PR does not add copy as of now and still in progress. Thanks for pointing about updating the docsy submodule. I'm not sure which terraform syntax highlighting are you referring to? Can you please clarify.

ankitm123:
For a WIP PR, you can add WIP to the title of the PR :)
Terraform codeblocks, for example: https://jenkins-x.io/docs/install-setup/create-cluster/eks/#cluster-provisioning

borntorock:
Updating the docsy theme submodule without enabling the `prism_syntax_highlighting` is making our code blocks blurred.

@ankitm123 Any idea / suggestions that might help resolve this. Looks like the updated javascript and css code in the docsy theme is overwriting the font of the code blocks, but not sure which one.

MarckK:
Looks good, @borntorock. Not blurry for me. 👍

borntorock:
> Looks good, @borntorock. Not blurry for me.

Yes, you're right. It was my browser's cache, displaying the blurred code blocks. Let me know check again after enabling the `prism_syntax_highlighting`.

borntorock:
the last build in this PR has the copy button on the code blocks, including terraform code blocks. Only thing is we have blurred code blocks but if we hover and preview the site in moon mode (black background), then the text in the code blocks is clearly visible. 

Looks due to some CSS over-ridden, but couldn't figure out which part of CSS is making this happen.

layouts/partials/head-css.html
```diff
@@ -12,3 +12,7 @@
 <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
 <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/highlight.js/9.15.10/styles/monokai.min.css">
 <link rel="stylesheet" href='{{ "css/style.css" | absURL }}' />
+{{ with .Site.Params.prism_syntax_highlighting }}
```
ankitm123, line 15: You can remove this, as this is already included in the docsy folder.

layouts/partials/scripts.html
```diff
@@ -83,4 +83,8 @@
 	}
 
 </script>
-{{ partial "hooks/body-end.html" . }}
\ No newline at end of file
+{{ with .Site.Params.prism_syntax_highlighting }}
```
ankitm123, line 86: Same, remove this as well. Don't need to include the prism css and js files

ankitm123:
@borntorock Left 2 comments.
The reason for the blurry text is a combination of these 2 lines: https://github.com/google/docsy/blob/master/static/css/prism.css#L11 and https://github.com/google/docsy/blob/master/static/css/prism.css#L13.

So, you can just copy the `prism.css` file into the static/css folder, and remove those 2 lines, and it will render properly.

Another thing, it would be nice if you can change the language from `javascript` to `terraform` [here](https://github.com/jenkins-x/jx-docs/blob/master/content/en/docs/_index.html#L56) and [here](https://github.com/jenkins-x/jx-docs/blob/master/content/en/docs/_index.html#L92).
Switching to prism from chroma makes them look bad (and also it's terraform and not js :D)

layouts/partials/scripts.html
borntorock: We do need to include the prism css and js files, as the prism js file is referenced in `layouts/partials/scripts.html` which is actually being over-ridden with the `scripts.html` we have in our jx-docs repository. Then, there's no reference to the prism js file. 

I did cross checked (after your comment) to see if it works or not but it didn't, by excluding the prism js file here.

borntorock:
Thanks for pointing out the suggestions. It does resolves the blurry text now. Also, we do need to reference the prism css and js files in `head-css.html` and `scripts.html` resp., more details [here](https://github.com/jenkins-x/jx-docs/pull/3057#discussion_r490723615).

> Another thing, it would be nice if you can change the language from `javascript` to `terraform` [here](https://github.com/jenkins-x/jx-docs/blob/master/content/en/docs/_index.html#L56) and [here](https://github.com/jenkins-x/jx-docs/blob/master/content/en/docs/_index.html#L92).

Done!

layouts/partials/head-css.html
borntorock: the prism css file also needs to be referenced but now from our static css folder similar to style.css.

borntorock:
@ankitm123 @MarckK the preview environment looks better with the clear text and copy button hover on the code blocks, so good to rebase and merge this PR. 

Let me know, if there's still something to add or change in this MR.

MarckK:
Nice work, @borntorock!

The only oddity with the code snippets is that some characters have grey highlighting. For example, [here](http://jx-docs.jx-jenkins-x-jx-docs-pr-3057.jenkins-x.live/docs/v3/guides/operator/):

Screenshot 2020-09-18 at 09 18 27

and [here](http://jx-docs.jx-jenkins-x-jx-docs-pr-3057.jenkins-x.live/docs/):
Screenshot 2020-09-18 at 09 21 53

ankitm123:
@MarckK that happens because we use the default theme. @borntorock I think if you use the prism css (I think the js one remains the same) file from the `prism-tomorrow` theme (This is the theme I use for my personal blog), it will fix the issue: https://prismjs.com/download.html#themes=prism-tomorrow&languages=markup+css+clike+javascript+bash+css-extras+docker+go+groovy+makefile+yaml&plugins=toolbar+copy-to-clipboard

ankitm123:
I think it looks good now @borntorock @MarckK. Time to get it merged? Any other issues?

PR #3058, opened 2020-09-16
Update decisions.md

biels:
draft.sh is deprecated and its repository is archived. It is confusing for newbies to find it here in the jenkins-x file. I believe its fucnctionality has been replaced by skaffold dev and skaffold debug. Correct me if I'm wrong.

### DIFF
--- a/content/en/docs/reference/components/decisions.md
+++ b/content/en/docs/reference/components/decisions.md
@@ -23,13 +23,13 @@ Jenkins X strongly recommends using public cloud managed Kubernetes clusters whe
 i.e. let folks that know how to run containers and manage clusters at scale so you can focus on adding value to your business.
 
 
-## Draft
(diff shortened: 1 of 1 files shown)


ankitm123:
@biels Draft is still used by JenkinsX (both in 2 and 3), but just for language detection and creating _build packs_. See: https://github.com/jenkins-x/jx-project/blob/60e2eca30202c84b462118faca4c202b26616236/go.mod#L6

biels:
I see. I didn't know that was the case. However, my intuition still tells me that since this is a deprecated package it should not be listed as a normal design decision here in the documentation. At least say that we are using the projecte even though it is deprecated because ... (because it still meets our needs for example)

ankitm123:
@jstrachan @rawlingsj thoughts?

PR #3076, opened 2020-09-25
fix: added how to install minikube for users of windows 10 home as hyper-v…

viveksahu26:
… is not supported in case of windows 10 home user

so ,one need to install it through docker driver.


viveksahu26:
@jstrachan  i have added the doc part which ,i asked you later on slack

PR #3080, opened 2020-09-28
Fix typo for jx step changelog documentation page

carlosrodlop:
typo


ankitm123:
@carlosrodlop You need to make this change in the jx codebase. The specific line: https://github.com/jenkins-x/jx/blob/master/pkg/cmd/step/step_changelog.go#L125

Issue #2473, opened 2020-03-01
[Request] New japanese translation


silverskyvicto:
I want to add a new Japanese translation to the document.

In this case, can I copy the existing English content together with the folder and add a Japanese translation?

sharepointoscar:
@silverskyvicto I think that is an awesome idea, and we would definitely like your contribution to this.

We are working on a restructure, and content refresh at the moment.  It is probably a good idea to wait just a bit, as there is also a PR #2674 which disables translations at the moment so that we first finish some work in the English version.

silverskyvicto:
@sharepointoscar 

The status of PR #2674 seems to be closed. What about the situation?

silverskyvicto:
What is the status of this?

ankitm123:
Majority of the focus has been on keeping the English docs updated, but if you want to work on this, that would be awesome!

silverskyvicto:
@ankitm123 

Is it OK to send a pull request according to the following documentation?

https://jenkins-x.io/community/documentation/

ankitm123:
> @ankitm123
> 
> Is it OK to send a pull request according to the following documentation?
> 
> https://jenkins-x.io/community/documentation/

Yes

silverskyvicto:
@ankitm123 

I submitted a pull request.

jenkins-x-bot-test:
@jenkins-x-bot: Closing this issue.

<details>

In response to [this](https://github.com/jenkins-x/jx-docs/issues/2473#issuecomment-737464779):

>Rotten issues close after 30d of inactivity.
>Reopen the issue with `/reopen`.
>Mark the issue as fresh with `/remove-lifecycle rotten`.
>Provide feedback via https://jenkins-x.io/community.
>/close


Instructions for interacting with me using PR comments are available [here](https://git.k8s.io/community/contributors/guide/pull-requests.md).  If you have questions or suggestions related to my behavior, please file an issue against the [jenkins-x/lighthouse](https://github.com/jenkins-x/lighthouse/issues/new?title=Command%20issue:) repository.
</details>

PR #3083, opened 2020-10-02
docs: First translate to 'ja' for index.html

silverskyvicto:
#2473 

I translated the first page into Japanese.


ankitm123:
@MarckK @jstrachan @rawlingsj thoughts?

ankitm123:
btw @silverskyvicto since we are moving to jx3, it would be awesome to look at translating jx3 (may not be now, but in the long term). https://jenkins-x.io/docs/v3/

silverskyvicto:
@ankitm123 

I haven't included it in the pull request, but I'm slowly working on translating the jx3 documentation.

MarckK:
Thank you for your PR @silverskyvicto. We are going to be undergoing a number of changes for the Jenkins X v3 release. Given this is the case, it may be better to resume translation in the coming months, towards 2021.

silverskyvicto:
@MarckK 

So the changes will be made.

I was translating the existing Jenkins X v3 documentation.
I'll think about the v2 one when v3 is all done.

msvticket:
It seems like this was abandoned. I don't see any point in merging something incomplete, that will only annoy any potential readers.

PR #3086, opened 2020-10-05
fix: first spike at changes doc

jstrachan:
and pipeline catalog docs


MarckK:
/lgtm 
Awesome, @jstrachan!

PR #3090, opened 2020-10-12
docs: how to change strategy for jx gitops upgarde

yelhouti:
Or maybe, we just replace this strategy here: https://github.com/jenkins-x/jx-gitops/blob/5194dc6e8628be69d56abf09cff86f47adacb979/pkg/cmd/kpt/update/update.go#L251
with `resource-merge`, unless @rawlingsj has a really good reason for doing this :p

### DIFF
--- a/content/en/docs/v3/guides/upgrade.md
+++ b/content/en/docs/v3/guides/upgrade.md
@@ -41,6 +41,14 @@ git push
 
 It is possible that you will have merge conflicts.  You can follow the inline git helper messages to resolve conflicts - or use your IDE to help figure out the merge issues. Usually if there are conflicts its safest to use the upstream version; particularly from the `versionStream` dir.
 
+Under the hood Kpt is used to fetch changes from the upstream defined in each Kptfile.  
+By default the `versionStream` directory has a Kpt strategy of `force-delete-replace` which removes all your changes in that folder. In order to merge your changes with the ones coming from upstream, add a file: `.jx/gitops/kpt-strategy.yaml` with the following content:
+```yaml
+config:
+  - relativePath: versionStream
+    strategy: resource-merge
+```
+
 Once ready, make a pull request onto your cluster repository, review changes and merge.  The [Jenkins X git operator](https://github.com/jenkins-x/jx-git-operator) will automatically apply the upgrades into your cluster.
 
 


rawlingsj:
Totally agree we should add some docs for this but yes there was a reason to use `force-delete-replace` because the version stream is meant to act more as an implementation detail used by Jenkins X to help reduce issues during upgrade. Changes can still be made by users outside of the version stream folder, I.e directly in a helmfile.yaml but changes inside the version stream can lead to changes complications firing upgrades.

yelhouti:
@rawlingsj I get it, but in that case why expose a folder that poeple are not allowed to touch or that get reset at each upgrade?
I tried with resource-merge and it worked pretty well. what is your concern exactly?

rawlingsj:
The reason the version stream is included inside the cluster git repo is so that we dont need to clone another source repo for the version stream when resolving versions.  This way the idea is that anyone can see the exact versions available in a release.

We could move the version stream out but then that can lead to more complexity.  If we document in the readme and the website under a customisation section that should help explain.  

> I tried with resource-merge and it worked pretty well. what is your concern exactly?

It depends on what customisations you made but we've seen merge conflicts on the version stream with `resource-merge`, so the concern is to set users up with a better chance of success during upgrades.  By keeping changes outside of the version stream then there is less risk users will have problems during an upgrade.  It's unclear how users might use customisations to the version stream so for now it's best to nudge folks away from doing that.

hervelemeur:
> it's best to nudge folks away from doing that

How about adding a README file at the root of versionStream folder to explain or link to this issue?

rawlingsj:
> How about adding a README file at the root of versionStream folder to explain or link to this issue?

Agreed - lets start with that and build up a customisations page on the website.

yelhouti:
@rawlingsj I don't think a merge failing is that big of a deal if kpt gives you the possibility to resolve conflicts. `jx gitops upgrade` will most definitely always be run by a user/dev and if he can't merge a conflict, well...
It also gives much more flexibility, allows the structure to stay clean by modifying yaml.gotml and doing quick dirty fixes when an upstream release introduces a bug... (which happened and I could override the change).
SO I'm still in favor of more flexibility by switching to merge strategy but you guys decide

rawlingsj:
Yeah its more helping folks down an easy path, we've learnt that just because something is technically possible it's not always the best, keeping changes out of the version stream seems the safest way to ensure upgrades work well with minimal manual interaction.  If you want to carry on using the merge strategy using the override its totally fine but for now it's probably best to recommend folks not do that.  I.e. favouring reliability, its hard to predict what customisations folks will do to a version stream if we don't advise against it which can lead to problems.

yelhouti:
You are probably right, but let's make sure that helmfile.yaml changes are not removed. this would allow a customization folder with all the extra values.yaml. It seemed to me that in some special case even helmfile was reset.

rawlingsj:
Yeah totally agree on that + using a seperate folder with an extra `values.yaml`, we need to make sure folks don't loose their changes to a `helmfile.yaml`.  FWIW `helmfile.yaml` isn't managed by `kpt` so that case sounds like a bug which we should fix, if you come across it again can you let me know please and let's find out why it happened.

hervelemeur:
Another place where it should be mentioned: last section of https://jenkins-x.io/docs/v3/guides/apps/#customising-charts

yelhouti:
> Another place where it should be mentioned: last section of https://jenkins-x.io/docs/v3/guides/apps/#customising-charts

indeed, I think it should only be there, I need to update this PR when I get the time...

hervelemeur:
> it should only be there

Maybe add a mention here too? https://jenkins-x.io/about/concepts/version-stream/

hervelemeur:
Related: #3095

PR #3507, opened 2022-01-05
docs: added note on github maven permissions

babadofar:
# Description
Added some info on the necessary permissions for using github as maven

### DIFF
--- a/content/en/v3/admin/setup/config/repository.md
+++ b/content/en/v3/admin/setup/config/repository.md
@@ -90,6 +90,8 @@ repositories:
     snapshotURL: https://maven.pkg.github.com/myowner/mysnapshots/
 ```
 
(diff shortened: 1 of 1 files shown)


content/en/v3/admin/setup/config/repository.md
```diff
@@ -90,6 +90,8 @@ repositories:
     snapshotURL: https://maven.pkg.github.com/myowner/mysnapshots/
 ```
 
+Note: The pipeline user will be used to push maven packages, so you may have to update the token for the pipeline user with additional permissions.
```
ankitm123, line 93: Can we add which additional permissions need to be added?

babadofar: sure

ankitm123: or link any documentation from github? that would help the users ...

PR #3546, opened 2022-02-19
fix: upgrade algolia docsearch to v3

babadofar:
Magically, this also seems to make search also work a little bit better on mobile 
image


static/css/style.css
```diff
@@ -62,3 +62,5 @@ a img {
         background-position: top center !important;
     } 
 }
+
+.DocSearch-Modal { top: 27px;}
```
ankitm123, line 66: NIT: new line

package.json
```diff
@@ -10,7 +10,7 @@
   },
   "dependencies": {
     "autoprefixer": "^9.8.5",
-    "docsearch.js": "^2.6.3",
+    "@docsearch/js": "^3.0.0-alpha.50",
```
ankitm123, line 13: alpha? No stable release yet?

babadofar: Coming soon? https://github.com/algolia/docsearch/issues/1303
"We are actually looking at open issues to see if we encounter any blockers for the stable v3 release. The next release ideally be the stable one!"

Can we get a PR env for this? I am in contact with Algolia support, they will help us set up the search, and public  access to a working environment would be very helpful

static/css/style.css
babadofar: Where? At the end?

layouts/partials/scripts.html
```diff
@@ -12,13 +12,14 @@
...
-    inputSelector: '#search-input',
-    algoliaOptions: { 'facetFilters': ["lang:{{ .Page.Language.Lang }}"] },
-    debug: false
-  });
+
+docsearch({
+  container: '#docsearch',
+  appId: 'GIN1OS4UKS',
+  indexName: 'jenkins-x',
+  apiKey: 'a9c99be0d65d1a8f6059be9104a12cd4',
```
ankitm123, line 20: Is this a search only api-key? https://www.algolia.com/doc/guides/security/api-keys/#search-only-api-key

static/css/style.css
ankitm123: yeah. github shows  a red circle when that's missing.

layouts/partials/scripts.html
babadofar: Yes it is

ankitm123:
@babadofar need to rebase with latest main branch.

PR #3577, opened 2022-03-29
fix: filter search results on language of current section

babadofar:


ankitm123:
From the preview url, seems like v2 results are still higher 🤔 
image

PR #3658, opened 2022-10-02
chore: blog for hacktoberfest 2022

ankitm123:

# Description

It's that time of the year!

### DIFF
--- a/content/en/blog/news/2022-hacktoberfest.md
+++ b/content/en/blog/news/2022-hacktoberfest.md
@@ -1,0 +1,49 @@
+---
+title: "Hacktoberfest 2022"
+date: 2022-10-03
+draft: false
+description: >
+  Contribute to Jenkins X as part of Hacktoberfest 2022
+categories: [blog]
+keywords: [Community, 2022]
+slug: "hacktoberfest2022"
+aliases: []
+author: Ankit D Mohapatra
+---
+
+We are excited to announce that Jenkins X will be participating in Hacktoberfest again this year!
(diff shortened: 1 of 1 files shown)


ankitm123:
We still need to label issues with hacktoberfest label, but cdf wants it by Monday for their blog post!

rajatgupta24 (approved):
🎉

ankitm123:
> So you should refer to these

Good point, adding them.

> * While issues mainly are placed in jenkins-x/jx the project is encompassed of many repositories in several organisation (you could refer to https://jenkins-x.io/v3/about/overview/source/)

Apart from this, I think a good thing to do with issues labelled for hacktoberfest is to make it clear where the code change can be? For example, if it's an issue related to jx project cli, it will most likely be in jx-project. It's not accurate (as it could be an issue in the dependencies like jx-helpers or go-scm), but can atleast make the life of the new contributor a lot simpler IMO. It's going to overwhelming for new contributors to look at the source documentation and figure out where the change needs to be made. 

Made some changes!

msvticket:
> I think a good thing to do with issues labelled for hacktoberfest is to make it clear where the code change can be? For example, if it's an issue related to jx project cli, it will most likely be in jx-project. It's not accurate (as it could be an issue in the dependencies like jx-helpers or go-scm), but can atleast make the life of the new contributor a lot simpler IMO. It's going to overwhelming for new contributors to look at the source documentation and figure out where the change needs to be made.

I agree

msvticket (approved):

PR #3674, opened 2022-11-13
fix: minikube install

babadofar:
Adding argument to  minikube to set the kubernetes version. Minikube uses 1.25 by default which we do not support yet.


rajatgupta24 (approved):
👍🏻

osamamagdy (approved):
LGTM

ankitm123 (approved):

